PR #70, opened 2016-10-09
Pancurses back-end

### LINKED ISSUE #69: pancurses back-end?
[`pancurses`](https://crates.io/crates/pancurses) uses either ncurses or pdcurses as back-end depending on the platform.

Since `ncurses-rs` doesn't work on Windowsx86_64 (https://github.com/jeaye/ncurses-rs/issues/78) this could make `cursive` actually build :P

I've had a cursory look at how the back-end system is implemented and am pretty sure this can be just simply feature-enabled if not completely switched out.

I can make a PR for this, just need a green light on one of the options.

nabijaczleweli:
To compile with `pancurses` run `cargo test --features=pancurses --no-default-features`

Closes #69


nabijaczleweli:
There are a couple of TODOs in `backend/curses/pan.rs` that need checking out.

nabijaczleweli:
This will also require `pancurses` to release current `HEAD`: https://github.com/ihalila/pancurses/issues/3

Cargo.toml
```diff
@@ -10,24 +10,35 @@ name = "cursive"
 readme = "Readme.md"
 repository = "https://github.com/gyscos/Cursive"
 version = "0.3.1"
+build = "build.rs"
 
 [build-dependencies]
 skeptic = "0.6"
 
+[features]
+cursive_ncurses = ["ncurses"]
```
nabijaczleweli: Probably because "Features and dependencies cannot have the same name: `ncurses`". Also, you'll need a `no_default_features` there :v

gyscos: The way it was before was working well. You don't need to _define_ the ncurses, pancurses or termion features, but you can still refer to them in the `default` array or in the `#[cfg]` directives.

src/backend/curses/pan.rs
```diff
@@ -0,0 +1,221 @@
...
+    fn poll_event(&self) -> Event {
+        // TODO: there seems to not be any indication
+        // of Ctrl/Alt/Shift in these :v
+        if let Some(ev) = self.window.getch() {
+            match ev {
+                pancurses::Input::Character(c) => Event::Char(c),
+                pancurses::Input::Unknown(i) => Event::Unknown(i),
+                // TODO: I honestly have no fucking idea what KeyCodeYes is
+                pancurses::Input::KeyCodeYes => Event::Refresh,
+                // TODO: There seems to not be a Key::Break??
```
gyscos: If you mean the "Pause, Break" key near "Print screen" and "Scroll Lock", it was not triggering any event on my terminal. But if it works for someone somewhere, I'll be glad to add it to the list of keys.

src/backend/curses/pan.rs
```diff
@@ -0,0 +1,221 @@
...
+                pancurses::Input::KeyBreak => Event::Refresh,
+                pancurses::Input::KeyDown => Event::Key(Key::Down),
+                pancurses::Input::KeyUp => Event::Key(Key::Up),
+                pancurses::Input::KeyLeft => Event::Key(Key::Left),
+                pancurses::Input::KeyRight => Event::Key(Key::Right),
+                pancurses::Input::KeyHome => Event::Key(Key::Home),
+                pancurses::Input::KeyBackspace => Event::Key(Key::Backspace),
+                // TODO: there seems to not be an F0 key for some reason?
```
gyscos: Is F0 a valid key? I never saw one, and it looked like a bitmask in the ncurses source, but I can add it if it's used.

nabijaczleweli: I guess it's like WinAPI's eternal compatibility :P

src/backend/curses/pan.rs
```diff
@@ -0,0 +1,221 @@
...
+                pancurses::Input::KeyF5 => Event::Key(Key::F5),
+                pancurses::Input::KeyF6 => Event::Key(Key::F6),
+                pancurses::Input::KeyF7 => Event::Key(Key::F7),
+                pancurses::Input::KeyF8 => Event::Key(Key::F8),
+                pancurses::Input::KeyF9 => Event::Key(Key::F9),
+                pancurses::Input::KeyF10 => Event::Key(Key::F10),
+                pancurses::Input::KeyF11 => Event::Key(Key::F11),
+                pancurses::Input::KeyF12 => Event::Key(Key::F12),
+                // TODO: there seems to not be any F-keys above F15
+                // for some reason?
```
nabijaczleweli: What should we do with this, then? Add F-keys up to F15 or wrap it to shift-F{1,2,3}?

src/backend/curses/pan.rs
```diff
@@ -0,0 +1,221 @@
...
+                pancurses::Input::KeyEIC => Event::Refresh,
+                pancurses::Input::KeyClear => Event::Refresh,
+                pancurses::Input::KeyEOS => Event::Refresh,
+                pancurses::Input::KeyEOL => Event::Refresh,
+                pancurses::Input::KeySF => Event::Refresh,
+                pancurses::Input::KeySR => Event::Refresh,
+                pancurses::Input::KeyNPage => Event::Refresh,
+                pancurses::Input::KeyPPage => Event::Refresh,
+                // TODO: no way to handle this
+                pancurses::Input::KeySTab => Event::Refresh,
```
gyscos: For this reason, the tab key is actually handled as `Key::Tab` rather then the `\t` char.

src/backend/curses/pan.rs
```diff
@@ -0,0 +1,221 @@
...
+                pancurses::Input::KeySResume => Event::Refresh,
+                pancurses::Input::KeySSave => Event::Refresh,
+                pancurses::Input::KeySSuspend => Event::Refresh,
+                pancurses::Input::KeySUndo => Event::Refresh,
+                pancurses::Input::KeySuspend => Event::Refresh,
+                pancurses::Input::KeyUndo => Event::Refresh,
+                pancurses::Input::KeyResize => Event::Refresh,
+                pancurses::Input::KeyEvent => Event::Refresh,
+                // TODO: not sure how to handle this
+                pancurses::Input::KeyMouse => Event::Refresh,
```
gyscos: Mouse support is currently absent, but sorta planned...

src/backend/curses/pan.rs
nabijaczleweli: Oops, missed it

src/backend/curses/pan.rs
nabijaczleweli: Addressed

src/backend/curses/pan.rs
nabijaczleweli: Added PauseBreak key

src/backend/curses/pan.rs
gyscos: We should try and see what pancurses returns when "Shift-F1" & cie are pressed. If F13 is returned in this condition, it should indeed map there.

src/backend/mod.rs
```diff
@@ -1,11 +1,13 @@
...
-#[cfg(feature = "ncurses")]
-pub mod curses;
 #[cfg(feature = "termion")]
-pub mod termion;
+mod termion;
+mod curses;
+
+pub use self::curses::*;
+#[cfg(feature = "termion")]
+pub use self::termion::*;
```
gyscos: I'd like to minimize the number of feature-conditional items. For instance, I'm thinking of having something like

```rust
#[cfg(feature = "termion")]
pub use self::termion::TermionBackend as Backend;

#[cfg(feature = "pancurses")]
pub use self::curses::PancursesBackend as Backend;
```
And removing the conditional from `src/lib.rs`.

nabijaczleweli: Now the back-end is at `backend::Concrete`, always.
`lib.rs` contains no more conditionals.

src/backend/curses/mod.rs
```diff
@@ -0,0 +1,40 @@
+use theme::{BaseColor, Color};
+
+#[cfg(feature = "ncurses")]
+mod n;
+#[cfg(feature = "ncurses")]
+pub use self::n::*;
```
gyscos: Any reason to use `*` instead of just `Concrete`?

src/backend/mod.rs
gyscos: :flushed:

gyscos: Looks like it's mostly requirements to build a dummy `Printer` for the doc tests, for instance `view/scroll.rs:139`... It should probably call `Concrete::init` instead of creating an empty struct. Doh, did I do that? :S

gyscos:
Note: ncurses-rs pushed a new version, breaking compatibility but kept the same major version. :(
`master` has been updated for that.

src/backend/curses/pan.rs
```diff
@@ -0,0 +1,219 @@
...
+        self.window.refresh();
+    }
+
+    fn print_at(&self, (x, y): (usize, usize), text: &str) {
+        self.window.mvaddstr(y as i32, x as i32, text);
+    }
+
+    fn poll_event(&self) -> Event {
+        // TODO: there seems to not be any indication
+        // of Ctrl/Alt/Shift in these :v
```
gyscos: For ncurses, a lot of the key+modifiers code were undocumented, outside of the usual code range, and I had to manually try every key combination, write down the bytes received by the application, and use this as mapping.

src/backend/curses/mod.rs
nabijaczleweli: DRY, basically

src/backend/curses/pan.rs
nabijaczleweli: For me it doesn't ever return F-keys, test exec:

```rust
extern crate pancurses;

fn main() {
    let pc = pancurses::initscr();
    loop {
        if let Some(ev) = pc.getch() {
            println!("{:?}", ev);
        }
    }
}
```

nabijaczleweli:
It builds!

src/backend/curses/pan.rs
gyscos: You may have to config the terminal emulator to let the Fn keys through.
When pressing F1 F2 F3 F4, I get:
```
Character('\u{1b}')
Character('O')
Character('P')
Character('\u{1b}')
Character('O')
Character('Q')
Character('\u{1b}')
Character('O')
Character('R')
Character('\u{1b}')
Character('O')
Character('S')
```

Similarly, other keys like `Ins`, `Del`, `Home`, ... also return blocks of 3 characters beginning with `'\u{1b}'`.

nabijaczleweli:
Should the `backend` module be made public? The tests seem to be failing because of that.

gyscos:
Doh, I forgot about that. I guess it can be made public but `#[doc(hidden)]`, I really don't want users to use it directly.

src/backend/curses/pan.rs
```diff
@@ -0,0 +1,218 @@
...
+                pancurses::Input::KeyF10 => Event::Key(Key::F10),
+                pancurses::Input::KeyF11 => Event::Key(Key::F11),
+                pancurses::Input::KeyF12 => Event::Key(Key::F12),
+                pancurses::Input::KeyF13 => Event::Shift(Key::F1),
+                pancurses::Input::KeyF14 => Event::Shift(Key::F2),
+                pancurses::Input::KeyF15 => Event::Shift(Key::F3),
+                // TODO: not sure what those are
+                pancurses::Input::KeyDL => Event::Refresh,
+                pancurses::Input::KeyIL => Event::Refresh,
+                pancurses::Input::KeyDC => Event::Refresh,
```
gyscos: `KeyDC` seems to correspond to the `Del` key.
`KeyIC` would be the `Insert` key.

src/backend/curses/pan.rs
```diff
@@ -0,0 +1,218 @@
...
+                pancurses::Input::KeyIC => Event::Refresh,
+                pancurses::Input::KeyEIC => Event::Refresh,
+                pancurses::Input::KeyClear => Event::Refresh,
+                pancurses::Input::KeyEOS => Event::Refresh,
+                pancurses::Input::KeyEOL => Event::Refresh,
+                pancurses::Input::KeySF => Event::Refresh,
+                pancurses::Input::KeySR => Event::Refresh,
+                pancurses::Input::KeyNPage => Event::Refresh,
+                pancurses::Input::KeyPPage => Event::Refresh,
```
gyscos: `KeyNPage` is apparently `PageDown`, and `KeyPPage` is `PageUp`

src/backend/curses/pan.rs
```diff
@@ -0,0 +1,218 @@
...
+                pancurses::Input::KeyEnter => Event::Key(Key::Enter),
+                // TODO: not sure what those are
+                pancurses::Input::KeySReset => Event::Refresh,
+                pancurses::Input::KeyReset => Event::Refresh,
+                pancurses::Input::KeyPrint => Event::Refresh,
+                pancurses::Input::KeyLL => Event::Refresh,
+                pancurses::Input::KeyAbort => Event::Refresh,
+                pancurses::Input::KeySHelp => Event::Refresh,
+                pancurses::Input::KeyLHelp => Event::Refresh,
+                pancurses::Input::KeyBTab => Event::Refresh,
```
gyscos: `KeyBTab` is sent when Shift-Tab is pressed.

src/backend/curses/pan.rs
```diff
@@ -0,0 +1,218 @@
...
+                pancurses::Input::KeySExit => Event::Refresh,
+                pancurses::Input::KeySFind => Event::Refresh,
+                pancurses::Input::KeySHome => Event::Refresh,
+                pancurses::Input::KeySIC => Event::Refresh,
+                pancurses::Input::KeySLeft => Event::Refresh,
+                pancurses::Input::KeySMessage => Event::Refresh,
+                pancurses::Input::KeySMove => Event::Refresh,
+                pancurses::Input::KeySNext => Event::Refresh,
+                pancurses::Input::KeySOptions => Event::Refresh,
+                pancurses::Input::KeySPrevious => Event::Refresh,
```
gyscos: `KeySPrevious` is `Shift-PageUp`.
`KeySNext` is `Shift-PageDown`.

gyscos: `KeySPrevious` is `Shift-PageUp`.
`KeySNext` is `Shift-PageDown`.

src/backend/curses/pan.rs
```diff
@@ -0,0 +1,218 @@
...
+                pancurses::Input::KeySCopy => Event::Refresh,
+                pancurses::Input::KeySCreate => Event::Refresh,
+                pancurses::Input::KeySDC => Event::Refresh,
+                pancurses::Input::KeySDL => Event::Refresh,
+                pancurses::Input::KeySelect => Event::Refresh,
+                pancurses::Input::KeySEnd => Event::Refresh,
+                pancurses::Input::KeySEOL => Event::Refresh,
+                pancurses::Input::KeySExit => Event::Refresh,
+                pancurses::Input::KeySFind => Event::Refresh,
+                pancurses::Input::KeySHome => Event::Refresh,
```
gyscos: `KeySHome` and `KeySEnd` are the Shift-version of Home and End.

src/backend/curses/pan.rs
```diff
@@ -0,0 +1,218 @@
...
+                pancurses::Input::KeyReplace => Event::Refresh,
+                pancurses::Input::KeyRestart => Event::Refresh,
+                pancurses::Input::KeyResume => Event::Refresh,
+                pancurses::Input::KeySave => Event::Refresh,
+                pancurses::Input::KeySBeg => Event::Refresh,
+                pancurses::Input::KeySCancel => Event::Refresh,
+                pancurses::Input::KeySCommand => Event::Refresh,
+                pancurses::Input::KeySCopy => Event::Refresh,
+                pancurses::Input::KeySCreate => Event::Refresh,
+                pancurses::Input::KeySDC => Event::Refresh,
```
gyscos: `KeySDC` is `Shift-Del`.

src/backend/curses/pan.rs
```diff
@@ -0,0 +1,218 @@
...
+                pancurses::Input::KeySDL => Event::Refresh,
+                pancurses::Input::KeySelect => Event::Refresh,
+                pancurses::Input::KeySEnd => Event::Refresh,
+                pancurses::Input::KeySEOL => Event::Refresh,
+                pancurses::Input::KeySExit => Event::Refresh,
+                pancurses::Input::KeySFind => Event::Refresh,
+                pancurses::Input::KeySHome => Event::Refresh,
+                pancurses::Input::KeySIC => Event::Refresh,
```
gyscos: I suppose `KeySIC` is `Shift-Ins`, but on my desktop it triggers "paste" instead.

src/backend/curses/pan.rs
```diff
@@ -0,0 +1,218 @@
...
+                pancurses::Input::KeySuspend => Event::Refresh,
+                pancurses::Input::KeyUndo => Event::Refresh,
+                pancurses::Input::KeyResize => Event::Refresh,
+                pancurses::Input::KeyEvent => Event::Refresh,
+                // TODO: mouse support
+                pancurses::Input::KeyMouse => Event::Refresh,
+                // TODO: not sure what those are
+                pancurses::Input::KeyA1 => Event::Refresh,
+                pancurses::Input::KeyA3 => Event::Refresh,
+                pancurses::Input::KeyB2 => Event::Refresh,
```
gyscos: `KeyB2` seems to be `Key::NumpadCenter`.
The other keys were probably the rest of the numpad before it got converted to Home/Pgup etc.

gyscos:
The compilation fails because `backend::Backend` needs to be imported in the doctests to provide `Concrete::init()`.
The termion build also fails, though I'm unsure why.

src/backend/termion.rs
```diff
@@ -1,20 +1,21 @@
+extern crate termion;
+
 use ::backend;
 use ::event::{Event, Key};
 use std::io::Write;
-use termion;
 use termion::input::TermRead;
```
gyscos: Should be `use self::termion::input::TermRead;` now that the crate is local.

gyscos:
Most doc-tests need to be `no_run`, since initializing a backend is not always possible without an actual terminal.

nabijaczleweli:
You'll need to either say which ones or make them `no_run` so I can rebase.

gyscos:
I'd say those failing:
```
    Cursive::add_global_callback_0
    Cursive::add_layer_0
    Cursive::find_0
    Cursive::find_id_0
    Cursive::menubar_0
    view::scroll::ScrollBase::draw_0
    views::edit_view::EditView_0
    views::select_view::SelectView_0
```

src/backend/curses/pan.rs
```diff
@@ -0,0 +1,218 @@
...
+        pancurses::init_pair(style.id(),
+                             find_closest(foreground) as i16,
+                             find_closest(background) as i16);
+    }
+
+    fn with_color<F: FnOnce()>(&self, color: ColorStyle, f: F) {
+        // TODO: pancurses doesn't have an `attr_get` equivalent
+        // let mut current_style: pancurses::attr_t = 0;
+        // let mut current_color: i16 = 0;
+        // pancurses::attr_get(&mut current_style, &mut current_color);
```
gyscos: Though well think about that after this PR, as the mechanism may go into `Printer` instead.

nabijaczleweli:
o boi, this seems to build now

gyscos:
Thanks!

nabijaczleweli:
Dope shit :+1:

PR #79, opened 2016-10-12
Fix warnings in back-ends

nabijaczleweli:
Also make BLT 80x40 because h=80 is fucking gigantic and doesn't fit on-screen, would be good to make it configurable somehow?

### DIFF
--- a/src/backend/blt.rs
+++ b/src/backend/blt.rs
@@ -3,10 +3,9 @@ extern crate bear_lib_terminal;
 use ::backend;
 use ::event::{Event, Key};
 use self::bear_lib_terminal::Color as BltColor;
-use self::bear_lib_terminal::geometry::{Point, Size};
+use self::bear_lib_terminal::geometry::Size;
 use self::bear_lib_terminal::terminal::{self, Event as BltEvent, KeyCode};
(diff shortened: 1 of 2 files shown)


gyscos:
I think the default terminal size is historically 80x24.
A configurable value would be very nice, but I'm not sure how to deal with backend specific configuration. The easiest way could be to try to read a hardcoded filename, but that's not very flexible. Maybe compile-time environment variables?...
Is like to avoid having backend-specific user code, but maybe it can't cleanly be done without.

src/backend/curses/mod.rs
```diff
@@ -11,6 +12,7 @@ mod pan;
 pub use self::pan::*;
 
 
+#[cfg(any(feature = "ncurses", feature = "pancurses"))]
```
gyscos: Maybe put those in `backend/mod.rs` instead? That way we'd only need one.

gyscos:
Making the window resizeable would already be a boost in user control actually.

src/backend/curses/mod.rs
nabijaczleweli: Addressed

PR #95, opened 2017-01-11
Added functions to register callbacks for ListView

Ameobea:
- Added `on_selected()` and `set_on_selected()` funtions that are called when the focus is changed
 - Added `focus()` to return the index of the currently focused item.

### DIFF
--- a/src/views/list_view.rs
+++ b/src/views/list_view.rs
@@ -1,9 +1,11 @@
+use Cursive;
 use Printer;
 use With;
 use direction;
-use event::{Event, EventResult, Key};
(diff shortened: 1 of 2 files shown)


gyscos:
I'm not sure whether we want the ID of the child in the callback in addition to the label. Let's go with this way for now and see if the need arises.

Ameobea:
My original plan was to expose the `Child`, but that would require changing some internals around and I wanted to leave as small an impact as necessary.

gyscos:
I'm not a huge fan of the current design of `ListView` anyway, so a refactor of the entire class one day™ may simplify that.

PR #96, opened 2017-01-11
Fixed issue with resetting background color for Termion backend

Ameobea:
Background color was being set back to the theme's background color rather than the terminal's background color when the backend was `finish()`ed.  This resets the background color to the default instead.


gyscos:
Thanks!

PR #104, opened 2017-01-30
Fixes issue #103

### LINKED ISSUE #103: Make chan-signal crate a dependency for Termion backend only
As far as i understand, the `chan-signal` crate is used for Termion backend only. However, it's listed in Cargo.toml as a _required_ dependency for the whole `cursive` crate, which prevents it to compile under Windows (because `chan-signal` is Unix only). So I've made a patch which makes `chan-signal` an optional dependency and creates a new `termion-backend` feature, which must be used by Cursive application instead of `termion` (I'm not sure that it's possible to "reuse" the name).
```diff
--- Cargo.toml	2017-01-24 09:53:59.000000000 +0300
+++ Cargo.toml	2017-01-30 13:20:32.774967600 +0300
@@ -22,12 +22,15 @@
 [dependencies]
 chan = "0.1.18"
 num = "0.1"
-chan-signal = "0.1"
 odds = "0.2"
 toml = "0.2"
 unicode-segmentation = "1.0"
 unicode-width = "0.1"
 
+[dependencies.chan-signal]
+optional = true
+version = "0.1"
+
 [dependencies.bear-lib-terminal]
 optional = true
 version = "1.3.1"
@@ -52,6 +55,7 @@
 
 [features]
 default = ["ncurses"]
+termion-backend = ["termion", "chan-signal"]
 
 [lib]
 name = "cursive"
```

hyst329:
So, I made a pull request that fixes #103. Now it builds and runs on Windows.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ script:
     - cargo build --verbose
     - cargo test --verbose
     -
-    - cargo build --verbose --features=pancurses --no-default-features
-    - cargo test --verbose --features=pancurses --no-default-features
+    - cargo build --verbose --features=pancurses-backend --no-default-features
+    - cargo test --verbose --features=pancurses-backend --no-default-features
     -
-    - cargo build --verbose --features=termion --no-default-features
-    - cargo test --verbose --features=termion --no-default-features
+    - cargo build --verbose --features=termion-backend --no-default-features
+    - cargo test --verbose --features=termion-backend --no-default-features
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ repository = "https://github.com/gyscos/Cursive"
 version = "0.4.0"
 
 [badges]
-travis-ci = { repository = "gyscos/cursive" }
+travis-ci = { repository = "gyscos/Cursive" }
 
 [build-dependencies.skeptic]
 optional = true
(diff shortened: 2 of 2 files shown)


Emilgardis:
Should this maybe be a `cfg(windows)` ?

hyst329:
@Emilgardis I don't think it should because there is no need for `chan-signal` when using other back-ends than Termion, even on other platforms.

gyscos:
* If we decide to use `*-backend` features, it should be done for the other backends as well.
* Using a target-dependant dependency is tempting as well, as it allows to keep using the `termion` as the feature name, though it's a bit of a hack.

Though this could be just the first of many issues related to backend-specific dependencies. I've avoided it until now, but it may not last.

hyst329:
@gyscos Why don't use `*-backend` names? They are maybe more informative (because such a name indicates that the feature represents a back-end and nothing other)

Emilgardis:
Travis is failing, https://travis-ci.org/gyscos/Cursive/jobs/196617575#L455

hyst329:
@Emilgardis Of course it will, because you're using `termion` feature instead of `termion-backend`

gyscos:
@hyst329 So can you update `.travis.yml` to use `--feature termion-backend` instead?

hyst329:
@gyscos Yes, I'll try to do it tomorrow, along with renaming all backend-related features (if it's necessary).

hyst329:
@gyscos It's OK now

PR #105, opened 2017-01-31
add shell.nix file for Nix users

dermetfan:
[NixOS](http://nixos.org/) users may need this file to build Cursive if they don't want to install ncurses in their user profile.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,10 @@
+{ pkgs ? import <nixpkgs> {} }:
+
+pkgs.stdenv.mkDerivation {
+  name = "cursive-env";
+  buildInputs = with pkgs; [
+    ncurses
+  ];
+
+  RUST_BACKTRACE = 1;
+}


gyscos:
Thank you for your contribution!

I don't know much about NixOS; when would this file be required?
Would this help if a NixOS user tries to compile a rust application that depends on cursive?

dermetfan:
> when would this file be required?

Every time they try to build Cursive. NixOS users enter a shell environment with the dependencies declared in `shell.nix` present so they get reproducible builds.

In theory those dependencies could also be installed system-wide, but I just tried that and it fails for some reason. I guess nix-shell environments keep references to build dependencies that are removed when the dependencies are installed as applications for the user.

> Would this help if a NixOS user tries to compile a rust application that depends on cursive?

In this case Cursive is fetched as a compiled library so `shell.nix` is not included. They would need to write their own `shell.nix` for that project. However, if we have a `shell.nix` here they can just copy our dependencies into theirs.

gyscos:
If nix users use cargo, I think it downloads dependencies and compiles them locally?

Anyway, I can merge that if there's cases where it helps.

dermetfan:
Even then, Cursive's `shell.nix` would not be downloaded into their project's root dir, so they'd need to type its full path somewhere in `target/build/cursive-*` every time. And if they need additional dependencies they still have to create their own file.

Cool, thanks!

Issue #90, opened 2016-12-06
Add text decorations


gyscos:
Allow a TextView to have special effects on parts of the content only: bold/italic/underline, different color/highlight...

dermetfan:
I [started](https://github.com/dermetfan/Cursive/commit/026eb9c28cfb5b2463e041e5b773323d2bd878bf) working on this. Maybe you can have a look so we agree on the direction to take this in?

You can test the current state by dropping the following in line 244 of `TextView` and running the `lorem` example. It just splits the text at "sit" and applies some style. (Yes, this is horrible and only works with the lorem ipsum text)
```rust
let delim = "sit";
let mut reappend_delim_i = 0;
let parts = text.split(delim)
    .map(|it| {
        let ret = if reappend_delim_i != 0 {
            format!("{}{}", delim, it)
        } else { it.to_owned() };
        reappend_delim_i += 1;
        ret
    })
    .collect::<Vec<_>>();
let parts = parts.iter()
    .map(|it| it.as_str())
    .collect::<Vec<_>>();
if parts.len() == 3 {
    let part1 = (parts[0], ::style::Style {
        color: None,
        ..Default::default()
    });
    let part2 = (parts[1], ::style::Style {
        color: Some(::theme::ColorStyle::Highlight),
        ..Default::default()
    });
    let part3 = (parts[2], ::style::Style {
        effect: ::theme::Effect::Reverse,
        ..Default::default()
    });
    printer.print_styled((x, 0), &[part1, part2, part3][..]);
} else {
    printer.print((x, 0), text);
}
```

gyscos:
I'd like, if possible, to separate the parsing part (where marked up text is parsed into a style tree) and the rendering part, to allow the same `TextView` to be easily filled from HTML, markdown or some other text format.

dermetfan:
I was planning to do parsing in its own module where we could implement different parsers for different markup languages.

I don't think we need a tree to store a parsing result. That is only useful if you want to modify the tree (like JavaScript inserting DOM nodes) but that's probably overkill so we can just parse the new markup string instead. The result could be stored as `Vec<StyledStr<'a>>` as you said in https://github.com/gyscos/Cursive/pull/106#commitcomment-20684775.

My current draft is

```rust
/// Handle parsing markup into style.
pub mod markup {
    /// Parses a markup string into `StyledStr`s.
    pub type Parse = for<'a> FnOnce(&'a str) -> Vec<StyledStr<'a>>;

[...]

PR #106, opened 2017-01-31
Text Style

dermetfan:
The functionality is there but we still lack any widgets using styled text and a possible markup language.

I'd suggest not to merge this PR yet but use it to track progress on this front. Next I'll try to implement styled text for `TextView` as per #90.

### DIFF
--- a/src/utils/styled_lines_iterator.rs
+++ b/src/utils/styled_lines_iterator.rs
@@ -1,0 +1,224 @@
+use utils::LinesIterator;
+use style::*;
+
+use unicode_width::UnicodeWidthStr;
+
+/// Generates rows of styled text in constrained width.
+///
+/// Given styled text and a width constraint, it iterates over
+/// collections of styled spans, each making up a row within the constraint.
+///
+/// # Examples
+///
+/// ```
+/// # use cursive::style::{StyledString, StyledText};
+/// # use cursive::utils::StyledLinesIterator;
+/// # let text_styled: &[StyledString] = &[].as_ref();
+/// let text_plain = text_styled.to_plain();
+/// let spans = StyledLinesIterator::make_spans(text_styled);
+/// // You can now keep `text_plain` and `spans` alive while using `iter`.
+/// let iter = StyledLinesIterator::new(text_plain.as_str(), spans.as_slice(), 10);
+/// ```
+pub struct StyledLinesIterator<'a> {
+    content: &'a [StyledSpan],
+    lines_iter: LinesIterator<'a>,
+
+    idx: usize,
+    offset: usize
+}
+
+impl<'a> StyledLinesIterator<'a> {
+    /// Returns the input as `StyledSpan`s so
+    /// you can keep them in scope when using a `StyledLinesIterator`.
+    pub fn make_spans(text: &'a [StyledString]) -> Vec<StyledSpan> {
+        let mut spans = Vec::with_capacity(text.len());
+
+        let mut cut = 0;
+        for part in text {
+            let (ref str, style) = *part;
+            let next_cut = cut + str.len();
+            spans.push(StyledSpan {
+                style: style,
+                start: cut,
+                end: next_cut,
+                width: str.width()
+            });
+            cut = next_cut;
+        }
+
+        spans
+    }
+
+    /// Returns a `StyledLinesIterator` over the given input.
+    ///
+    /// Get the `plain` text from [`StyledText::to_plain()`](../style/trait.StyledText.html#tymethod.to_plain) and the `styled_spans` from `make_spans()`.
+    pub fn new(plain: &'a str, styled_spans: &'a [StyledSpan], width: usize) -> Self {
+        StyledLinesIterator {
+            content: styled_spans,
+            lines_iter: LinesIterator::new(plain, width),
+            idx: 0,
+            offset: 0
+        }
+    }
+
+    /// See [`LinesIterator::show_spaces()`](struct.LinesIterator.html#method.show_spaces).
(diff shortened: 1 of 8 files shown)


src/utils/styled_lines_iterator.rs
```diff
@@ -0,0 +1,224 @@
...
+///
+/// # Examples
+///
+/// ```
+/// # use cursive::style::{StyledString, StyledText};
+/// # use cursive::utils::StyledLinesIterator;
+/// # let text_styled: &[StyledString] = &[].as_ref();
+/// let text_plain = text_styled.to_plain();
+/// let spans = StyledLinesIterator::make_spans(text_styled);
+/// // You can now keep `text_plain` and `spans` alive while using `iter`.
```
dermetfan: I don't really like this but couldn't figure out a way to avoid it.

We need to keep the string that the `LinesIterator` uses alive since it is not owned by the `StyledLinesIterator`.

Because the parsed `StyledStrings` use `Cows` now and not simply slices, the user also needs to join them together into `text_plain` at some point so he can use `StyledSpan`s against it later.

gyscos: The problem is that we mix direct indices in a string (`StyledSpan`) and `&str` (`StyledString`), and it's not easy to move from one to the other.

Since we can now print content which is not directly present in the input (because of escaping), the indices from `StyledSpan` are no longer a great solution. Keeping a `Cow` in the `StyledSpan` sounds easier.
Actually, I think we could merge `StyledSpan` and `StyledString`. Both refer to a "styled" section from the input text, potentially "rewritten" to account for escaping (hence the `Cow`).

Then, we'd have:

```rust
fn parse(input: &'a str) -> Vec<StyledSpan<'a>>
fn compute_rows(spans: &'b [StyledSpan<'a>]) -> Vec<StyledSpan<'a>>
```

For the `compute_rows`, I'm thinking of updating the existing `LinesIterator` rather than wrapping it. I'm not a big fan of `StyledLinesIterator` taking both the plain text and the spans, and hoping that they'll match. Especially since it doesn't _need_ the plain text.

I'll try to work on updating `LinesIterator` for that this week if I find some time.

dermetfan: Excuse me, I have a little trouble understanding your idea.

> Since we can now print content which is not directly present in the input (because of escaping), the indices from StyledSpan are no longer a great solution. Keeping a Cow in the StyledSpan sounds easier.
> Actually, I think we could merge StyledSpan and StyledString. Both refer to a "styled" section from the input text, potentially "rewritten" to account for escaping (hence the Cow).

If the indices are removed, `StyledSpan` contains only the `style` and `width`. Since `StyledString` already contains both of them (`width` via `styled_string.0.as_ref().width()`) merging them would just mean that `StyledSpan` vanishes.

`compute_rows` is basically equivalent to `styled_lines_iter.collect()`, it just solves the ownership problem of `StyledLinesIterator` by removing the ability to iterate, keeping the ownership in the parent scope.

However I also don't like the potential for misuse that `StyledLinesIterator` poses.

Therefore I would suggest to do perform the m stillerge of `StyledSpan` and `StyledString`, in effect just removing `StyledSpan`, and making `StyledLinesIterator` split `StyledString`s in half on line breaks. Since it now uses `Cow` that should be possible. We gain:

- no more potential misuse of `StyledLinesIterator`
- no more numerical indices tricking the borrow checker
- no `text_plain` to be kept alive by the user, it now "lives" in the `Cow`s of a `[StyledString]`

I'll try this so we can see how it compares with your approach. Maybe I missed something, especially since I'm not sure exactly what you mean.

gyscos: While updating `LinesIterator`, I found one of the reasons I was using indices: when inserting text in a `TextArea`, it tries not to recompute rows for the entire text, and just "offset" the indices in most situations.

Now, this is mostly an optimisation (so removing it should be slower, but safe), and `TextArea` may end up using an entirely different structure anyway.

gyscos: Regarding the `compute_rows` signature I posted, it is indeed overly simplistic, and was just meant to show an example of input/output flow.

Though thinking about it again, we may need 2 structures after all: the width only needs to be computed when creating the rows, so the input for the updated `LineIterator` could be a list of

```rust
struct StyledStr<'a> {
    content: Cow<'a, str>,
    style: Style
}
```
and the output would use:
```rust
struct StyledSpan<'a> {
    content: Cow<'a, str>,
    style: Style,
    width: usize,
}
```
We'll need the width multiple times and don't want to compute it each time, but we don't need it (and should not compute it) before computing the rows. Having a single type would be a bit ugly, since the `width` field would just be left empty by the parser, and would be ignored by the `LineIterator`, so having two types makes sense.

gyscos: Actually, we don't need `Cow` at all for `StyledSpan`, since it *always* points to the `StyledStr`.

gyscos: Also, owning_ref doesn't look like it would be enough, be [refstruct-rs](https://github.com/diwic/refstruct-rs) looks very promising.

Virtuos86:
I need very strong this TextView functionality. When it can be implemented?

gyscos:
I'm currently a bit busy with work & family. Should have some more time in october ; mouse support and styled text will be the two focus points then.

gyscos:
Styled text is now implemented.

A basic markdown parser exist, and more may come, but the core issue is resolved.
Thanks for helping with the design!

Virtuos86:
Good news! Thanks  a lot!

PR #107, opened 2017-02-01
return length and width from utils::prefix(_length)

dermetfan:
I considered using a tuple but that is not as readable. Also I hope you find it OK to rename `prefix_length` to `prefix` as I think that makes more sense now that it returns a `Prefix` that contains not only the length.

see df835d1a10442fd243f8cd85c305b1451a17475f

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -9,7 +9,15 @@ mod reader;
 pub use self::lines_iterator::{LinesIterator, Row};
 pub use self::reader::ProgressReader;
 
-/// Computes the length of a prefix that fits in the given `width`.
+/// The length and width of a part of a string.
+pub struct Prefix {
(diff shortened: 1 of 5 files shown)


src/utils/mod.rs
```diff
@@ -34,7 +42,7 @@ pub use self::reader::ProgressReader;
 /// prefix_length(my_text.graphemes(true), 5, "");
```
gyscos: Just need to update the doc comment here and it should be fine.
Thanks!

dermetfan: Sorry, I need to remember that `cargo build` may not be enough

gyscos: That's what travis is for :)

gyscos:
Looks good! Thanks again!

PR #110, opened 2017-03-05
Correct colors on pancurses+Windows and use strongly typed attributes

ihalila:
I've just released a new version of pancurses, which in turn is using an updated version of PDCurses on Windows. Earlier PDCurses defaulted to BGR colors, which caused blue and red to be flipped in Cursive because you use the numbers directly that ncurses does. I now build PDCurses with RGB as I think it's a little bit less surprising that way.

So with the update to pancurses 0.8, Cursive dialogs now look correct on Windows when building against the pancurses backend.

While I was at it I also changed the pancurses backend to use the Attribute and ColorPair types instead of using chtypes directly, as the API is going to become more strongly typed and deprecate the use of just numbers everywhere. Also Window now has the attrget() function so pan::Concrete doesn't need to track the current ColorStyle separately.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ version = "5.85.0"
 [dependencies.pancurses]
 features = ["wide"]
 optional = true
-version = "0.7"
+version = "0.8"
 
 [dependencies.termion]
 optional = true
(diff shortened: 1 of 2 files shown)


gyscos:
Looks good to me!
Thanks again! I had no idea colors were inverted. Was the problem windows-only? I could run `examples/theme.rs` on linux, and the `ncurses` and `pancurses` backends are giving similar results.

ihalila:
Yeah it only happened on Windows, PDCurses defaults to different colors (so 1 was blue instead of red, and so forth).

Issue #113, opened 2017-03-06
prefix function off


ConnyOnny:
```rust
utils::prefix("abra a".split(' '), 5, " ").length
```
is 6 because of a difficult-to-spot bug where the delimiter length is not added everywhere it needs to be. PR incoming in a few minutes.

PR #114, opened 2017-03-06
fix utils::prefix bug: #113

ConnyOnny:


src/utils/mod.rs
```diff
@@ -66,7 +68,9 @@ pub fn prefix<'a, I>(iter: I, available_width: usize, delimiter: &str) -> Prefix
 
     // We counted delimiter once too many times,
     // but only if the iterator was non empty.
-    let length = if sum == 0 { sum } else { sum - delimiter_len };
+    let length = sum.saturating_sub(delimiter_len);
+
+    debug_assert!(current_width <= available_width);
```
ConnyOnny: The `debug_assert` would have helped to more quickly track down the error. The `saturating_sub` may be faster, is safer (if for some reason 0<sum<delimiter_length, the previous version would panic) and I think it also looks nicer. It already has a good comment, so it's hard not to understand that line anyway.

gyscos:
Looks good!
Last thing: could you add a test with the input from #113? Thanks!

gyscos:
Doh, turns out this actually is wrong when the input begins with an empty token (for instance a string that begins with a space).
30cac851e7cd61de2b2df5de9b7251b00c4fe059 should fix this.

PR #115, opened 2017-03-06
use Iterator::sum() for summing instead of fold

ConnyOnny:
Just makes it a little more idiomatic. Nothing important.


gyscos:
Looks good, thanks!
I think I wrote that when sum was still unstable, and forgot to come back to it.

PR #127, opened 2017-04-02
EditView: Add a customizable filler character

afonso360:
This PR allows the user to select a custom filler character on EditView

Turns out converting from char to &str is not that easy, and we can't store the &str without adding a lifetime, if you know of a better way to solve this please tell me.


gyscos:
I think using a `String` as a filler would make sense: some "characters" are made of multiple codepoints.
Also, some chars can be double-width (most CJK characters), in which case we need to print half as much characters.

gyscos:
Thanks!

PR #134, opened 2017-04-21
Automatically grow textarea

untitaker:
This shouldn't break any usecases since currently everybody uses `BoxedViews`
for the `TextArea` to look any good.

### DIFF
--- a/src/views/text_area.rs
+++ b/src/views/text_area.rs
@@ -13,8 +13,9 @@ use view::{ScrollBase, SizeCache, View};
 
 /// Multi-lines text editor.
 ///
-/// A `TextArea` will attempt to grow vertically dependent on the content.
-/// Wrap it in a `BoxView` to constrain its size.
+/// A `TextArea` will attempt to grow vertically and horizontally
+/// dependent on the content.  Wrap it in a `BoxView` to
+/// constrain its size.
 pub struct TextArea {
     // TODO: use a smarter data structure (rope?)
     content: String,


src/views/text_area.rs
```diff
@@ -362,6 +362,11 @@ impl TextArea {
 }
 
 impl View for TextArea {
+    fn required_size(&mut self, constraint: Vec2) -> Vec2 {
+        self.compute_rows(constraint);
+        Vec2::new(1, self.rows.len())
+    }
```
gyscos: We could use the longest row as width here instead of 1?

untitaker: I fixed this, but that doesn't really work as expected. The problem is that `compute_rows` computes the rows dependent on the `constraint`, so it may wrap unnecessarily.

Should I pass something like `Vec2::new(MAXSIZE, MAXSIZE)` to `compute_rows` to fix this or write my own linecounting?

src/views/text_area.rs
```diff
@@ -362,6 +363,14 @@ impl TextArea {
 }
 
 impl View for TextArea {
+    fn required_size(&mut self, constraint: Vec2) -> Vec2 {
+        self.compute_rows(constraint);
+        Vec2::new(
+            self.rows.iter().map(|r| r.width).max().unwrap_or(1),
```
gyscos: I think it needs a `+1` here to work properly (we need 1 extra column for the cursor itself).

untitaker:
Any comment on my concern in the old review?

On 22 April 2017 01:12:59 GMT+02:00, Alexandre Bury <notifications@github.com> wrote:
>gyscos commented on this pull request.
>
>
>
>> @@ -362,6 +363,14 @@ impl TextArea {
> }
> 
> impl View for TextArea {
>+    fn required_size(&mut self, constraint: Vec2) -> Vec2 {
>+        self.compute_rows(constraint);
>+        Vec2::new(
>+            self.rows.iter().map(|r| r.width).max().unwrap_or(1),
>
>I think it needs a `+1` here to work properly (we need 1 extra column
>for the cursor itself).
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/gyscos/Cursive/pull/134#pullrequestreview-34139374

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

src/views/text_area.rs
gyscos: If the constraint sets a maximum width, then the line wraps are justified, aren't they?

gyscos:
I tried locally, and after adding 1 to the columns as I commented, it seems to be working quite well.

untitaker:
Sorry for slow reply and thanks for merging!

On 22 April 2017 06:52:12 GMT+02:00, Alexandre Bury <notifications@github.com> wrote:
>Merged #134.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/gyscos/Cursive/pull/134#event-1053318431

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #136, opened 2017-04-26
Add termion AlternateScreen support

quininer:
This PR make `termion` consistent with other backend behavior.


gyscos:
Looks good, thanks!

PR #137, opened 2017-05-10
added splice feature (rust issue #32310)

ZOTTCE:
Now splice is unstable feature
https://github.com/rust-lang/rust/issues/32310

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -57,6 +57,7 @@
 //! and log to it instead of stdout.
 //!
 //! Or you can use gdb as usual.
+#![feature(splice)]
 #![deny(missing_docs)]
 
 extern crate toml;


gyscos:
Adding this feature would prevent this library from building on stable, which is a problem.

But we're not using `Vec::splice` here, we're using the method from [`odd::vec::VecExt`](https://docs.rs/odds/0.2.25/odds/vec/trait.VecExt.html).

ZOTTCE:
Yeap, my bad, didn't see last commits.
Cursive doesn't work from cargo, that's problem (at least yesterday).

gyscos:
Ah just understood this wasn't released yet.
Just published 0.5.1 which should be working better.

PR #139, opened 2017-05-18
The change will allow for programs to easily use Cursive in place of regular prtinln! and similar functions.

someonesapien:
Allows for programs to have utils like one given below. Allows for easy use of cursive.

[cursive_util.rs.txt](https://github.com/gyscos/Cursive/files/1009898/cursive_util.rs.txt)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -587,7 +587,7 @@ impl Cursive {
     /// [`step(&mut self)`]: #method.step
     /// [`quit(&mut self)`]: #method.quit
     pub fn run(&mut self) {
-        self.run = true;
+        self.running = true;
         // And the big event loop begins!
         while self.running {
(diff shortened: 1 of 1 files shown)


gyscos:
Thanks for the report!

I'm not sure I understand everything here. Could you precise the use- case, and why we need these modifications?

gyscos:
As of 900bdfea5835a67ad3465ec644b72984a9a2ea59, `Cursive::run` can be called multiple times if you want to re-use an existing view tree for another event loop.

Regarding the other changes, I'm not sure making those fields public is the best solution.

gyscos:
I'll close this for now, as I think the main point (call Cursive multiple times) is now solved. Feel free to re-open if I'm wrong.

PR #152, opened 2017-08-23
Use `log` crate

hcpl:
This provides a uniform way to handle logging and can be combined with logs from other libraries in end-user apps.

Also removes dependency on `eprintln!()` macro which is stable only recently, in Rust 1.19.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ optional = true
 version = "0.10"
 
 [dependencies]
+log = "0.3"
 num = "0.1"
 odds = "0.2"
 owning_ref = "0.3"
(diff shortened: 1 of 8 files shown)


gyscos:
I feel like in this case, it isn't logging as much as debugging. In normal behavior, it should *never* print anything on `stderr`. The only exceptions are panics, and people debugging cursive itself (temporary print statements that shouldn't stay in the release). As such, since those statements are supposed to stay commented out in the repository, I don't think they matter much.

I agree though that the `unreachable!` call with message is much better than the `eprintln/unreachable` combo. Ideally not a single `eprintln!` is uncommented in the code, removing the requirement for a recent rust compiler.

hcpl:
I can think of a situation where debug-level logging output is useful: when one of the maintainers requests it from the crate user to ease the investigation.

Also personally I often use debug-level logging statements for libraries. Since they can be easily switched off, they don't interfere with the normal code flow, while providing a way to help us out in critical situations.

gyscos:
> I can think of a situation where debug-level logging output is useful: when one of the maintainers requests it from the crate user to ease the investigation.

Ah, good idea indeed.

PR #155, opened 2017-10-11
fix some clippy warnings

KlossPeter:


gyscos:
Thanks a lot! I have been ignoring clippy a bit lately...

Regarding `OnEdit`/`OnSubmit`, is it fine to use them in the public API without re-exporting them (from the `views` module`)?

PR #156, opened 2017-10-26
Ignore keys properly in SelectView::on_event

adrianN:
on_event consumes all keypresses by trying to jump to the item that matches the char. Even if no such item is found, the key event is lost to the rest of the application. By returning EventResult::Ignored when the key press has no effect, the key event that have no effect on the SelectView can still be used by somebody else.

### DIFF
--- a/src/views/select_view.rs
+++ b/src/views/select_view.rs
@@ -437,6 +437,8 @@ impl<T: 'static> SelectView<T> {
                     // Apply modulo in case we have a hit
                     // from the chained iterator
                     self.focus.set(i % self.items.len());
+                } else {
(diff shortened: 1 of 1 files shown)


gyscos:
Good idea!

Thanks a lot!

TerminalWitchcraft:
This issue still persists. Running `cargo run --example select` on tag 0.9.0 and hitting Capital letter(Shift + key) on keyboard jumps to the item in the list. 

This issue came up when I created a select_view, which had an item whose name starts with letter `q`, I couldn't exit the program, eventhough I had `q` mapped to `siv.quit()` as global_callback. Tested with ncurses, as well as termion backend.

Issue #165, opened 2017-12-08
Suggestion: StackView::pop() returns the popped layer


alekratz:
My use case is that I have two separate logging windows that I want to flip back and forth between, in the same position on-screen. I figured I'd implement this by doing `StackView::pop()` twice, and push the popped values back in the same order they were popped. However, I can't (easily) do this since `StackView::pop()` doesn't return anything.

Since a `StackView` is just backed by a `Vec`, it would be a trivial change to have it return some `Option<Layer>`, right? What do you think about this?

gyscos:
That's a good idea! Just a few points:
* This means `pop_layer` now returns something, so it may confuse closures that were only calling that. Though one solution is to only modify the `StackView` method and not the shortcut one defined on `Cursive` (which is what you did in the PR).
* The returned view is a `Box<View>`. It now *does* implement `View`, so it can be given as-is to `add_layer`, but it means it'll get a second layer of boxing. This may not be such a problem (especially if you don't do it too often). Possible solutions would be:
  * Use something like a `Into<Box<View>>` trait to accept both direct views and boxed views, so it's only boxed once.
  * Add a separate method `add_boxed_layer(&mut self, Box<View>)`

alekratz:
>This means pop_layer now returns something, so it may confuse closures that were only calling that. 

Yeah, I had to fix something in `Dialog` as a result of this. It's probably going to be a breaking change for some people.

>Though one solution is to only modify the StackView method and not the shortcut one defined on Cursive (which is what you did in the PR).

Ah, I didn't see `pop_layer` was also being called in `Cursive` as well. Personally, I think it'd be more consistent if it returned the popped item - but that is up to you. Making that change cascades down into `MenuPopup`, so that's going to need brackets around it, too.

IMO, if you do accept the PR, you may want to hold off until the next minor version because of the breaking changes.

>The returned view is a Box<View>. It now does implement View, so it can be given as-is to add_layer, but it means it'll get a second layer of boxing. This may not be such a problem (especially if you don't do it too often).

That's a good point, I wasn't aware that `Box<View>` implemented `View`.

>Use something like a Into<Box<View>> trait to accept both direct views and boxed views, so it's only boxed once.

Were you thinking of adding that trait bound to `add_layer_at`? That's where the view actually ends up being boxed. I think I'll try that first.

>Add a separate method add_boxed_layer(&mut self, Box<View>)

[...]

PR #166, opened 2017-12-08
Have StackView::pop() return the item that was popped off

alekratz:
This is the implementation of my suggestion in #165. Let me know what you think.

### DIFF
--- a/examples/text_area.rs
+++ b/examples/text_area.rs
@@ -41,8 +41,8 @@ fn main() {
                         ).unwrap();
                         find(s, &text);
                     })
-                    .button("Cancel", |s| s.pop_layer()),
-            ).on_event(Event::Key(Key::Esc), |s| s.pop_layer()),
+                    .button("Cancel", |s| { s.pop_layer(); }),
+            ).on_event(Event::Key(Key::Esc), |s| { s.pop_layer(); }),
         )
     });
 
--- a/src/cursive.rs
+++ b/src/cursive.rs
@@ -393,10 +393,12 @@ impl Cursive {
         self.screen_mut().add_fullscreen_layer(view);
     }
 
-    /// Convenient method to remove a layer from the current screen.
-    pub fn pop_layer(&mut self) {
-        self.screen_mut().pop_layer();
+    /// Convenient method to remove a layer from the current screen. The removed layer is returned,
+    /// if there was any.
+    pub fn pop_layer(&mut self) -> Option<Box<View>> {
+        let popped = self.screen_mut().pop_layer();
         self.clear();
+        popped
     }
 
     // Handles a key event when it was ignored by the current view
--- a/src/views/dialog.rs
+++ b/src/views/dialog.rs
@@ -154,7 +154,7 @@ impl Dialog {
 
     /// Shortcut method to add a button that will dismiss the dialog.
     pub fn dismiss_button<S: Into<String>>(self, label: S) -> Self {
-        self.button(label, |s| s.screen_mut().pop_layer())
+        self.button(label, |s| { s.screen_mut().pop_layer(); })
     }
 
     /// Sets the title of the dialog.
--- a/src/views/menu_popup.rs
+++ b/src/views/menu_popup.rs
@@ -164,7 +164,7 @@ impl MenuPopup {
                             action_cb.clone()(s);
                         }
                     }),
-                ).on_event(Key::Left, |s| s.pop_layer()),
+                ).on_event(Key::Left, |s| { s.pop_layer(); }),
             );
         })
     }
--- a/src/views/stack_view.rs
+++ b/src/views/stack_view.rs
@@ -106,14 +106,15 @@ impl StackView {
     }
 
     /// Adds a view on top of the stack.
-    pub fn add_layer_at<T>(&mut self, position: Position, view: T)
+    pub fn add_layer_at<V, T>(&mut self, position: Position, view: T)
     where
-        T: 'static + View,
+        V: View + 'static,
+        T: Into<Box<V>>,
     {
         self.layers.push(Child {
             // Skip padding for absolute/parent-placed views
             view: Box::new(
-                ShadowView::new(Layer::new(view))
+                ShadowView::new(Layer::new(*view.into()))
                     .top_padding(position.y == Offset::Center)
                     .left_padding(position.x == Offset::Center),
             ),
@@ -133,10 +134,10 @@ impl StackView {
         self.with(|s| s.add_layer_at(position, view))
     }
 
-    /// Remove the top-most layer.
+    /// Attempts to remove the top-most layer, returning it if there was one.
     pub fn pop_layer(&mut self) -> Option<Box<View>> {
         self.layers.pop()
-            .map(|child| child.view)
+            .map(|child| { child.view })
     }
 
     /// Computes the offset of the current top view.
--- a/src/views/stack_view.rs
+++ b/src/views/stack_view.rs
@@ -134,8 +134,9 @@ impl StackView {
     }
 
     /// Remove the top-most layer.
-    pub fn pop_layer(&mut self) {
-        self.layers.pop();
+    pub fn pop_layer(&mut self) -> Option<Box<View>> {
+        self.layers.pop()
+            .map(|child| child.view)
     }
 
     /// Computes the offset of the current top view.


gyscos:
Good idea - and thank you very much!

Could you just update the documentation for the `pop_layer` method?

Thanks again!

alekratz:
I just realized I hadn't checked the examples, and they're failing because of these changes. I'll need to get those fixed up, too.

gyscos:
The examples use `Cursive::pop_layer`, which hasn't been changed yet. I think they are compiled during `cargo test`, which is run on each PR, so you're good for now :)
Though updating `Cursive::pop_layer` will indeed require some modifications there.

gyscos:
Using `Into<Box<View>>` itself may prove a bit difficult:
* We need `impl <V: View> Into<Box<View> for V`, which conflicts with `impl <T, U: From<T>> Into<U> for T`
* We can't define `impl <V: View> From<V> for Box<View>` since it conflicts with `impl<T> From<T> for T`

Even if we use a custom IntoBoxedView trait, we would need specialization anyway, since `Box<View> implements `View`.

alekratz:
It may fail to compile with non-default features? I didn't test those.

gyscos:
Now I realize the pop'ed view will actually contain a `ShadowView<Layer<V>>` (or sometimes just a `Layer<V>`), which we may not want to re-wrap in a second layer and shadow view. 

Ideally the pop_layer would unwrap the `V` from there. Or alternately a method would allow to add a "raw" view, without any wrapper.

It starts to sound a little more complex than the original use-case warrants. Retrieving pop'ed views is still a nice-to-have features, but if the goal is to switch the order of views in a `StackView`, there may be easier ways (like `StackView::bring_to_front()` or similar methods).

gyscos:
Another problem I just found: we're not getting the correct implementation of `Into<Box<V>>` when giving a `Box<View>`. That's because `View` is unsized, so doesn't match `V: View + 'static`.

This means `Box<View> != Box<V>`, so we don't get `impl <T> From<T> for T`. Instead, `Box<View>` is treated as a `View` and re-boxed from `impl <T> From<T> for Box<T>`.

alekratz:
I've been thinking about this a lot this weekend and I think my current solution, as it stands, would be pretty unsafe overall. Even if we solve the double-boxing issue, consider this scenario

1. I push some View to a StackView. For example purposes, let's say it's a TextView.
2. That TextView gets wrapped in a Layer, which gets wrapped in a ShadowView.
3. I later go to pop that TextView off, expecting a TextView. Instead, I get a ShadowView, and the downcast attempt fails and panics because of it. That's surprising, and definitely violates the principle of least astonishment.

Not to mention that pushing a freshly-popped object will wrap it another Layer and ShadowView. I don't think this is the right approach to this problem. I'll think about better, safer solutions for this kind of thing. I still think that pop_layer should return the top View, but there's gotta be a better way to do that than what I have right now.

gyscos:
Adding `unwrap` to the `ViewWrapper` is a great idea, and I'm surprised it's not there already.

But as is, the `StackView` only has a `Vec<Box<View>>` - it doesn't even know it has a `ViewWrappers`.

One solution could be to change `StackView` to instead store `Vec<ShadowView<Layer<Box<View>>>>`, and configure `ShadowView` to enable/disable the shadows through a variable. This way, the `StackView` can call `unwrap` twice as you mentioned.

Unfortunately, it's not trivial to get the static type of the view from there: the content of the children doesn't have the `Any` bound (we can't have a Box trait object with two bounds: `Box<View + Any>` is invalid), so the best we can do currently is use `call_on_any` to access the view itself. We don't know if the view even has an ID, so we'd need to use the (non-yet implemented) view path with an empty path (indicating the root view) to select it. This would give us a `&mut` access to the view, on which we could call `std::mem::replace` to extract the actual view. Phew!

The last few steps could be made a bit easier if we had `View: Any`. I'm thinking of adding that; this would mean any type implementing `View` would need to be `'static`, but since we currently have a `'static` bound on `StackView::add_layer`, this is virtually already the case. This does close the door to my dream of eventually removing this `'static` bound, but maybe it's for the best.

alekratz:
I must admit - this is pushing the limits of my knowledge of the Rust type system. I don't mind learning, as long as you don't mind teaching.

alekratz:
Wow, thanks for the verbosity. I think I understand. Let me see if I get this right, correct me where I'm wrong:

* `View`s aren't `Any` (well, now they are, but they weren't.) This means we can't (couldn't) call `Box::downcast` on any `Box<View>`.
* Now that `View` is bound by `Any`, we *can* do the above. 🎉🎉🎉
* Since `StackView` stores its children either as `ShadowView<Layer<T>>` *or* `Layer<T>` (depending on whether shadow is needed), this adds a layer of annoyance that we need to work around. Your suggestion is to add a field to determine whether the `ShadowView` is actually active or not, so we can always have the same wrapping of `ShadowView<Layer<T>>` - whether shadows are active or not.
    * Assuming my understanding is correct, as an aside I'd say this is the point where we'd want to introduce a new private type, or have the `Child` structure handle this discrepancy (rather than adding an `active` field to the `ShadowView` struct - it feels redundant).
* Since we can't verify at compile-time that the popped-off item is going to be of some `T`, we use `Box::downcast` to either return `Some(T)` (with the `and_then` chain), statically typed, because `Box<T> where T: Any` is magic.

I'm going to look at the source some more, and probably close this PR and start over when I feel like I've got something I can send you - I didn't make many changes, no skin off my back. Thanks so much for the write-up, it helped me understand the situation a lot better.

gyscos:
> * Views aren't Any (well, now they are, but they weren't.) This means we can't (couldn't) call Box::downcast on any Box<View>.
> * Now that View is bound by Any, we can do the above. 🎉🎉🎉

Correct!

> * Since StackView stores its children either as ShadowView<Layer<T>> or Layer<T> (depending on whether shadow is needed), this adds a layer of annoyance that we need to work around. Your suggestion is to add a field to determine whether the ShadowView is actually active or not, so we can always have the same wrapping of ShadowView<Layer<T>> - whether shadows are active or not.
>     * Assuming my understanding is correct, as an aside I'd say this is the point where we'd want to introduce a new private type, or have the Child structure handle this discrepancy (rather than adding an active field to the ShadowView struct - it feels redundant).

Indeed, a cleaner solution would be to have a private enum that could be either `ShadowView<Layer<T>>` or `Layer<T>`.

> *Since we can't verify at compile-time that the popped-off item is going to be of some T, we use Box::downcast to either return Some(T) (with the and_then chain), statically typed, because Box<T> where T: Any is magic.

Actually, the `and_then` chain here is mostly due to the fact the `ViewWrapper::into_inner` returns a `Result` (since it *could* fail to return the inner view, in some rare cases).

We don't actually *need* to downcast the view here; we could stop right before, return the unwrapped `Box<View>`, and let the user downcast if needed. But I feel it's much nicer for the users not to have to downcast it themselves.

Argh, but I hoped users could still call it with `T = View`, so that they could still get a `Box<View>` if they didn't know the actual type, but it turns out they can't - we can only downcast to a concrete type, not to a trait object, even if it's the *very same trait we are starting from*... :(

Oh well, I guess we'll have to return the `Box<View>` itself then, and either let users downcast it themselves, or add another function to downcast it for them (depending on how common a use-case it really is).

gyscos:
In commit 1b7170e3084ba8408566f231840165e93eb59184, StackView implement this enum solution to store either a `ShadowView<Layer<Box<View>>>` or a `Layer<BoxView<T>>`, and in commit bd9b6a695b739815872a10982d27149002e6e061 `pop_layer` return this `Box<View>`.

Next thing will be find a way to not re-box a `Box<View>` when adding a new layer.

gyscos:
Let's consider the use-case of pop'ing a view, and putting it back (possibly later), repeated multiple times:
```rust
cursive.add_layer(Dialog::new());

for i in 0..20 {
    let layer = cursive.pop_layer();
    cursive.add_layer(layer);
}
```
The target is to avoid resulting in a large chain of nested boxes.

One solution is to add a dedicated method to add a boxed view: this is easy to implement, but starts to crown the API.
The other solution is to find some polymorphism to allow `add_layer` to behave differently if given a `Dialog` (where it needs to be boxed) and a `Box<View>` (where it should not be boxed). One problem being that `Box<View>` implements `View`, so technically it would always be valid to re-box it. What we want it some specialization to skip this step when not necessary - and as the wording suggests, it looks like the specialization feature would make this easier (or even possible at all).

`StackView::add_layer_at` needs a `Box<View>`. It currently takes a `<T: View>` and box it in a `Box<View>` as its first instruction.

So what we need is a trait to result in a `Box<View>`, implemented for `<T: View>`, and with a specialization for `Box<View>` (or maybe for `Box<T: ?Sized + View>`?).

This sounds a lot like `Into<Box<View>>`. The problem is that this is *not* implemented for `T: View`, so we wouldn't be able to pass `Dialog::new()` directly.
Two solution from there: find a way to implement `Into<Box<View>>` for `T: View`, or create our own trait.

* Trying directly `impl <T: View> Into<Box<View>> for T` fails with a conflicting implementation, as the standard library already provides `impl<T, U: From<T>> Into<U> for T`.
* Trying `impl <T: View> From<T> for Box<View>` conflicts with `impl<T> From<T> for T`
* Creating our own `IntoBoxedView` trait runs into the same conflicting implementation issue, since `Box<View>` implements `View`.

Specialization would at least enable the third point, and maybe the others, but it's not here yet.

Another solution is to do this check at runtime, by exploiting the fact that `View: Any`: we can check if `view.is::<Box<View>>`, but then I'm not sure how to actually convert `view` into the `Box<View>`, as it seems `Any` only allow downcasting references. :S

gyscos:
Note: we couldn't actually use `Box::downcast`, as it's only defined for `Box<Any>`, not `Box<AnyView>`; and it wasn't easy to get a `Box<Any>` from a `Box<AnyView>`.

So I just added `AnyView::as_boxed_any` to solve this part: we can now actually downcast a `Box<AnyView>` to a `Box<TextView>`.

gyscos:
I think I found a solution!

First, the whole `View`/`AnyView` is now simpler: instead of `AnyView: View`, we now have `View: AnyView` (any `AnyView` is still auto-implemented for any `T: View`). It feels like a cyclic type dependency, but it works, so champagne! This means we only have to worry about `Box<View>` now, not `Box<AnyView>`.

Second, I added a `IntoBoxedView` trait, and implemented it on `Box<View>` and any `T: View`. It's like `Into<Box<View>>`, but it actually works for `T: View`. `StackView::add_layer` and similar methods now use this for a trait bound instead of `T: View`, which means a `Box<View>` can be taken as argument and isn't re-boxed.

In order to have this implementation and to avoid specialization (which isn't stable yet), I had to remove `impl View for Box<View>`. This means `Box<View>` is no longer considered a `View`. This isn't a problem for `add_layer` since we use `IntoBoxedView` instead, but it does mean `Box<View>` cannot be used as-is in some view wrappers like `Layer`. For those (less common) cases, I added a `ViewBox` type, which is really just a wrapper around a `Box<View>` and implements `View`.

gyscos:
This is all in master, and should finally resolve the problem addressed by this PR.

PR #168, opened 2017-12-11
Add rotate_up() and rotate_down() methods to StackView

alekratz:
Since the idea of returning the currently popped view is not a trivial implementation I figured I'd add `rotate_up` and `rotate_down` methods for modifying which item is currently on top of the stack. This may warrant a use case for a `VecDeque` instead of a `Vec`, but that's up to you.

### DIFF
--- a/src/views/stack_view.rs
+++ b/src/views/stack_view.rs
@@ -138,6 +138,19 @@ impl StackView {
         self.layers.pop();
     }
 
+    /// Moves the bottom layer of the stack to be on the top.
+    pub fn rotate_down(&mut self) {
+        let item = self.layers.remove(0);
+        self.layers.push(item);
+    }
+
+    /// Moves the top layer of the stack to be on the bottom.
+    pub fn rotate_up(&mut self) {
+        let item = self.layers.pop()
+            .expect("rotate_up expects there to be at least one item on top of the StackView");
+        self.layers.insert(0, item);
+    }
+
     /// Computes the offset of the current top view.
     pub fn offset(&self) -> Vec2 {
         let mut previous = Vec2::zero();


gyscos:
Thank you very much for this PR!

I think `rotate_up`/`rotate_down` methods have a few problems:
* The term `rotate` itself might be confusing
* It doesn't allow every view permutation we may want. Given 3 views, it's impossible to only bring the middle one to the front, leaving the other two intact.

I think I'd prefer a `push_to_back(source: usize)`/`bring_to_front(source: usize)` taking a layer id as argument, and/or maybe a `move_to(source: usize, target: usize)` for more precise insertion (though 1) it's a less common use-case, and 2) it can be simulated by a few calls to `push_to_back` or `bring_to_front`).

The main problem is that the layer ID may not be obvious (is it starting from the top? from the bottom?), and having to query the number of views to get the index of the last view is annoying. 
* One solution to that would be to index based on the method used: `push_to_back` would start from the top, and `bring_to_front` would start from the bottom. This way, pushing the front view to the back and bringing the back view to the front (the two most common uses, I would imagine) can both take `0` as a layer number. The big problem here, obviously, is the lack of coherence between the two methods.
* Another solution is to have a new enum as index:

  
```rust
enum LayerPosition {
    FromBack(usize),
    FromFront(usize),
}
```

This would make `push_to_back`, `bring_to_front` and `move_to` less ambiguous, and a bit easier to write - especially with shortcut methods like `LayerPosition::top() -> Self { LayerPosition::FromFront(0) }`.

gyscos:
Another use-case, not yet covered by this solution, would be to find the layer containing a certain view (through its ID for instance), and bring that one to front. I had an idea of a `ViewPath` describing the position of a view in the layout tree, and, while I don't have anything precise in mind, I think it could be used to find the layer position, and give this value to `bring_to_front`.

gyscos:
Commit 8da659367d55a2a7b0746af0225ddc6b291036c8 brings `StackView::{move_layer,move_to_front,move_to_back}` - does this look like it could solve your use-case?

gyscos:
With `StackView::find_view_from_id`, we can now find the layer containing a target view.
`ViewPath` may still be intereting down the line, but isn't necessary for now.

Issue #167, opened 2017-12-11
Titleposition in Dialogview


punkkeks:
A cool feature would be to able to specify the position of the title in a dialogview. 
You can e.g. write 

    DialogView::new("Foo").title_pos(Align::top_left())...

which would result in 
![grafik](https://user-images.githubusercontent.com/921462/33841376-dadf0fcc-de97-11e7-914c-88aedf683c93.png)

of course center would be an illegal position and there is a question how to deal with it. I think the easiest solution would be a `panic!()`, but a compiletime error would be better. 

Just drop your thoughts about this here and I will try to implement this :)

gyscos:
Good idea!

It could take a `HAlign` to only care about horizontal alignment. Having the title on the bottom doesn't sound a common use-case to warrant to possible panic.

punkkeks:
Hmm, what about vertical dialog titles? So you could possibly "rotate" the text by 90° and have it on the vertical borders of the box? It would certainly a fun idea and we would cover all usecases (except) for the center one) 

Am 11. Dezember 2017 18:12:06 MEZ schrieb Alexandre Bury <notifications@github.com>:

gyscos:
Hmm I'm not sure how to rotate text yet (it'd be a bit more complex since characters are not square, but twice as tall as they are wide).

I think for now, a `HAlign` would be the simplest option, and would match the way buttons are aligned. If/when different layouts for title and buttons are added, we'll think of a more general design.

punkkeks:
Let me clarify the rotation part. Instead of writing

    text

You write 

    t
    e
    x
    t

But yes, I see the problems, which arise and I agree, that top halign would be the simplest method. Let's see, what I can do today



Am 11. Dezember 2017 18:31:10 MEZ schrieb Alexandre Bury <notifications@github.com>:

punkkeks:
Another point to mention is, how big the distance between the border and the actual text should be? Do you have a policy for that, e.g. fixed vs percentage? 



Am 11. Dezember 2017 18:31:10 MEZ schrieb Alexandre Bury <notifications@github.com>:

gyscos:
I would have put all the way to either direction:

```
┌┤ Title ├────┐
│             │
```

```
┌────┤ Title ├┐
│             │
```

As for vertical text, it could fit well for some japanese/chinese text, but paradoxically it'd be a pain for those, since each character would be 2-columns wide, and wouldn't fit well on the border.

gyscos:
Thank you for the PR!

PR #169, opened 2017-12-12
added option to position the header in a dialog

punkkeks:
Added the functionality to define the position of the title. I "verified" it by changing the dialog example to procude this:

grafik

### DIFF
--- a/src/views/dialog.rs
+++ b/src/views/dialog.rs
@@ -362,11 +362,8 @@ impl Dialog {
             if len + 4 > printer.size.x {
                 return;
             }
-            let x = match self.title_position {
-                HAlign::Left => 3,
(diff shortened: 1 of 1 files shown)


punkkeks:
This will solve #167

punkkeks:
Indeed... Looks like my implementation, okay :) changed it

gyscos:
Looks great!

Last tiny detail: could you mention in the doc that it defaults to center?

Thanks again!

punkkeks:
Where exactly? In the struct doc, method doc?

gyscos:
I think in the method doc is the easiest solution, no need to describe every initial state in the struct doc itself.

gyscos:
Alright, perfect!

Issue #182, opened 2018-01-08
Fullscreen dialog in pancurses on Windows


agavrilov:
Bug report

# Problem description

I create a full screen dialog using add_fullscreen_layer on Cursive and full_screen method on Dialog. When I resize the window created by pancurses, the dialog is not resized with the window. It keeps its size unchanged.

Sample code where you can see an issue:
```rust
    let mut app = Cursive::new();
    app.add_fullscreen_layer(
        Dialog::text("Sample")
        .button("Cancel", move |s| {
            s.quit();
        })
        .full_screen()
    );
    app.run();
```

# Environment

* Operating system: windows
* Backend used: pancurses
* Current locale: English
* Cursive version: 0.7.5 from crates.io

gyscos:
Blast! It seems we fail to detect window resizing. Apparently it's not reported the way it is on Linux... :S

Anyway, thanks for the report!

gyscos:
I suppose this is fixed? Thanks for the PR! :D

PR #183, opened 2018-01-09
Call resize_term on KeySize

agavrilov:
Fix for #182.  We need to call resize_term(0, 0) to adjust pancurses internal structures. See [here](https://github.com/ihalila/pancurses#terminal-resizing)


gyscos:
Thank you very much!

PR #190, opened 2018-01-22
Add get_content on Dialog

xliiv:


gyscos:
Thanks for the PR!
This looks like it should work fine; how are you trying to use it?

xliiv:
I added broken test, because I'm not sure how to cast the view.
If I find out I'll fix it.
I'm not sure where is the code which gives error from the issue desc. so I can't paste it.

xliiv:
@gyscos 
I don't know what i'm doing wrong.
For me, this should work.
`None` is returned when type is wrong (https://doc.rust-lang.org/std/any/trait.Any.html#method.downcast_ref)

gyscos:
Ah yes, I think it's something tricky: `Box<View>` implements `View`. So the current `get_content` actually returns a `&Box<View> as &View`, rather than dereferencing the box.
It could be fixed by returning `&*self.content.view` instead.

xliiv:
It's `Box` indeed, so it should be dereferecned, i'm stupid or tired or both :)
How do you know that `Box<View>` implements `View`?
I did `rg "impl View"|rg Box` in `Cursive` source and got 0 hits.

gyscos:
It's a bit convoluted, but there is a line

```
impl <U, T> ViewWrapper for T
    where T: Deref<Target=U> + DerefMut + 'static,
          U: View + ?Sized
{
    // ...
}
```

So any type that can dereference to a `View` implements ViewWrapper on that view, and therefore `View`.

gyscos:
Note: I removed the need for the `view_any!()` macro, so no need to fix the `mines` example anymore.

xliiv:
Removed

gyscos:
Looks good, thanks!

xliiv:
I've got this code
```rust
// dynamically adds view
let widget = field.get_widget(); // returns Box::new(Checkbox::new()) or other boxed views
list_view.add_child(field_name.as_ref(), widget);

// .. cut

for (idx, field) in self.fields.iter().enumerate() {
    if let &ListChild::Row(ref label, ref view) = list_view.get_row(idx) {
        // delegate getting value from a widget
        let value: String = get_value(&**view);
        eprintln!("field, value: {:?}, {:?}", label, value);
    } else {
        eprintln!("DELIMITER");
    }
}

// .. cut

fn get_value(&self, view: &AnyView) -> String {
    let value = view.as_any().downcast_ref::<Checkbox>().unwrap().is_checked();
    format!("{}", value)
}
```

Sorry for pasting it here, but I think It's similar case to [this](https://github.com/gyscos/Cursive/pull/190#issuecomment-359533271) 
Help me @gyscos  with this (again :( ), please and I'll solve an issue as a payback :)

gyscos:
So the problem here may be that there's no specific method to add a boxed view, and without specialization, this means adding a `Box<CheckBox>` will end up treating `Box<CheckBox>` as a `View` and storing a `Box<Box<CheckBox>>` (and this is what you'll get in the `ListChild::Row`).

For now, you can downcast it

xliiv:
I've got problem with downcasting. Because I can't downcast this variant `Box<View>` as a `View`.
So Should I downcast it like this:
`let value = view.as_any().downcast_ref::<Box<Box<Checkbox>>>().unwrap().is_checked();`
?
Btw, this doens't work either. :(

gyscos:
Glad you found the solution!
I wish there was an easier way, or at least a way to debug why you don't get the view you're looking for. :-/

If you're looking for something to work on, first - thank you so much! I just added #192 - it could be a good way to start. ;)

Thanks again!

xliiv:
Unfortunately, the "solution" was a mess in my debug code (i overwrote actual Box<View> with debugging specific View, like Checkbox so it worked, but real case still gives error :().
I decided to skip it.. and do as much as i can (while skipping it), publish sources and then ask for help referring the source code.

Ad. #192 I'll take it tomorrow, BTW i can't assign myself.

PR #194, opened 2018-01-25
Allow getting child in parent views

### LINKED ISSUE #192: Add a way to access child view in view parents
Many views wrap one (or many) child view, yet offer no way to access it directly.

This includes:

* `SelectView`: no method to access an item
* Most wrapper views (`BoxView`, `Panel`, ...)

Ideally, we would have const and mutable accessor for each child.

xliiv:
Fixes #192

### DIFF
--- a/src/view/view_wrapper.rs
+++ b/src/view/view_wrapper.rs
@@ -219,7 +219,7 @@ macro_rules! wrap_impl {
 /// }
 ///
 /// impl<T: View> FooView<T> {
-///     inner_getters!(T);
+///     inner_getters!(self.view: T);
 /// }
 ///
 /// impl <T: View> ViewWrapper for FooView<T> {
@@ -229,7 +229,7 @@ macro_rules! wrap_impl {
 /// ```
 #[macro_export]
 macro_rules! inner_getters {
-    ($t:ty) => {
+    (self.$v:ident: $t:ty) => {
         /// Gets access to the inner view.
         pub fn get_inner(&self) -> &$t {
             &self.view
--- a/src/view/view_wrapper.rs
+++ b/src/view/view_wrapper.rs
@@ -39,6 +39,9 @@ pub trait ViewWrapper: 'static {
     where
         F: FnOnce(&mut Self::V) -> R;
 
+    /// Gets the reference to the wrapped view.
+    fn get_view(&self) -> &Self::V;
+
     /// Attempts to retrieve the inner view.
     fn into_inner(self) -> Result<Self::V, Self>
     where
@@ -119,6 +122,10 @@ where
     {
         Some(f(self.deref_mut()))
     }
+
+    fn get_view(&self) -> &Self::V {
+        self.deref()
+    }
 }
 
 // The main point of implementing ViewWrapper is to have View for free.
@@ -196,8 +203,8 @@ macro_rules! wrap_impl {
             Some(f(&mut self.$v))
         }
 
-        fn into_inner(self) -> Result<Self::V, Self> where Self::V: Sized {
-            Ok(self.$v)
+        fn get_view(&self) -> &Self::V {
+            &self.$v
         }
     };
 }
--- a/src/view/view_wrapper.rs
+++ b/src/view/view_wrapper.rs
@@ -119,7 +119,6 @@ where
     {
         Some(f(self.deref_mut()))
     }
-
 }
 
 // The main point of implementing ViewWrapper is to have View for free.
--- a/src/view/view_wrapper.rs
+++ b/src/view/view_wrapper.rs
@@ -39,9 +39,6 @@ pub trait ViewWrapper: 'static {
     where
         F: FnOnce(&mut Self::V) -> R;
 
-    /// Gets the reference to the wrapped view.
-    fn get_view(&self) -> &Self::V;
-
     /// Attempts to retrieve the inner view.
     fn into_inner(self) -> Result<Self::V, Self>
     where
@@ -123,9 +120,6 @@ where
         Some(f(self.deref_mut()))
     }
 
-    fn get_view(&self) -> &Self::V {
-        self.deref()
-    }
 }
 
 // The main point of implementing ViewWrapper is to have View for free.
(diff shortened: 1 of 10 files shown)


xliiv:
I'm not sure about my concept but this issue looks like two steps:
1. add `get_view` (and `get_view_mut`) for `ViewWrapper` which will handle:
    *  https://docs.rs/cursive/0.7.5/cursive/views/struct.IdView.html
    * https://docs.rs/cursive/0.7.5/cursive/views/struct.Layer.html
    * https://docs.rs/cursive/0.7.5/cursive/views/struct.OnEventView.html
    * https://docs.rs/cursive/0.7.5/cursive/views/struct.Panel.html
    * https://docs.rs/cursive/0.7.5/cursive/views/struct.ShadowView.html
    * https://docs.rs/cursive/0.7.5/cursive/views/struct.SizedView.html
    * https://docs.rs/cursive/0.7.5/cursive/views/struct.TrackedView.html
    * https://docs.rs/cursive/0.7.5/cursive/views/struct.BoxView.html

2. add to each "container view", like:
    * [ ] https://docs.rs/cursive/0.7.5/cursive/views/struct.SelectView.html
    * [ ] https://docs.rs/cursive/0.7.5/cursive/views/struct.ListView.html
    * [ ] https://docs.rs/cursive/0.7.5/cursive/views/struct.LinearLayout.html
    * [x] https://docs.rs/cursive/0.7.5/cursive/views/struct.Dialog.html (done before this PR)

Ad 1: I've got [an issue](https://github.com/gyscos/Cursive/pull/194/files#diff-d49f0ab04beeb6460a2f3e0a60556196R103) which I'm struggling with
Ad 2: It looks easy, and requires only of adding similar code snippet to each of 4 View?

gyscos:
Thanks for the good work!

Ah yes, `IdView` is a bit tricky and cannot simply return a reference to the parent view (since there may be some runtime checks like a `Rc`). It's fine not to have a getter for this one - since it's an `IdView`, the child can always be retrieved anyway.

xliiv:
I'm a bit confused so let's sum it up this ticket:
1. Add `get_child`/`get_child_mut` to `SelectView`
2. Add `get_child`/`get_child_mut` individually for all of the `ViewWrapper` except `IdView`"
Is this correct?

Which name should I use, because each view has own name :) ?

gyscos:
This sounds good.
As for the naming, it's true they're not quite uniform. :-/

I would say:
* For: `SelectView` (just remembered it doesn't actually store child views, but items):
    * To match `item`/`add_item`, a `get_item(&self, i: usize) -> Option<(&str, &T)>` 
    * Maybe a `get_item_mut(&mut self, i: usize) -> Option<(&mut String, &mut T)>` using `Rc::get_mut`?
* For the `ViewWrapper` views except `IdView`, add `get_child` and `get_child_mut` (let's use the same name for every view).
    * The name could alternately be `get_inner`/`get_inner_mut`. This would mark the difference between parent views (with multiple children and `get_child` method) and wrapper views (with a single view and `get_inner`). What do you think?

src/views/box_view.rs
```diff
@@ -263,4 +265,25 @@ mod tests {
         assert_eq!(Vec2::new(10, 1), full.required_size(Vec2::new(10, 1)));
         assert_eq!(Vec2::new(10, 10), full.required_size(Vec2::new(10, 10)));
     }
+
+    #[test]
+    fn test_get_inner() {
```
xliiv: I'm not sure this tests should stay.
I can remove it or change.

src/views/on_event_view.rs
```diff
@@ -203,6 +203,16 @@ impl<T: View> OnEventView<T> {
             },
         );
     }
+
+    /// Gets access to the inner view.
+    pub fn get_inner(&self) -> &T {
+        &self.inner
```
xliiv: This is not `inner_getters!` becuase view here is defined as:
`self.inner`
and in the rest of `ViewWrapper` it's
`self.view`

xliiv:
Although I did it like in PR, common sense tells me that design should be changed more like this:
* current `ViewWrapper` is `ViewWrapper` and `IdView` implements it
* some new trait is introduced `ProperViewWrapper2Name`  which inherits from `ViewWrapper` and  defines `get_inner(_mut)` 

However I'm not sure if it's easy possible or possible

xliiv:
I think it's ready. What do you think?

src/views/on_event_view.rs
gyscos: Let's rename it then! Either that or name all the others `inner`. No real preference here, but renaming it to `view` means fewer changes.

gyscos:
> Although I did it like in PR, common sense tells me that design should be changed more like this:
> * current ViewWrapper is ViewWrapper and IdView implements it
> * some new trait is introduced ProperViewWrapper2Name which inherits from ViewWrapper and defines get_inner(_mut)
> However I'm not sure if the juice is worth the squeeze :)

I was thinking that too, with `ViewWrapper` implemented for any `T: ProperViewWrapper2Name`. Thinking about it more, `ProperViewWrapper2Name` is pretty much just `Deref + DerefMut`, and we already implement `ViewWrapper` for these types - so maybe we could have all such "wrappers" simply derefs into their inner type? I don't know if this makes sense.

Otherwise, no need to overthink this too much. Pushing it to a custom traits makes some things more elegant, but also means we need to import this trait to access the inner view, which is a small con.

src/views/box_view.rs
gyscos: Tests are never bad!

src/views/on_event_view.rs
xliiv: I've created seperate PR: https://github.com/gyscos/Cursive/pull/195

xliiv:
Btw:
is this intentional? and if what's the meaning?
https://github.com/gyscos/Cursive/blob/master/src/views/tracked_view.rs#L15
https://github.com/gyscos/Cursive/blob/master/src/views/tracked_view.rs#L22

gyscos:
Hmmm nope, looks like I just added a block forgetting the other one existed? <_<

gyscos:
> so maybe we could have all such "wrappers" simply derefs into their inner type? I don't know if this makes sense.

Scratch that, we often want to tune the ViewWrapper implementation (for instance `BoxView` changes ths `required_size` method, ...), so the simple implementation we'd get from `Deref + DerefMut` wouldn't be enough.

gyscos:
Giving it a second look, it may be cleaner to add the variable name in the macro: just like `wrap_impl!(self.view: T)`, it would be `inner_getters!(self.view: T)` . This way there's a little less magic ; when seeing the macro call, we have a better idea it uses `self.view` internally.

gyscos:
Perfect, thanks again!

xliiv:
No. Thank you for such a great lib :)

src/view/view_wrapper.rs
```diff
@@ -201,3 +201,42 @@ macro_rules! wrap_impl {
...
+///     wrap_impl!(self.view: T);
+/// }
+/// # fn main() { }
+/// ```
+#[macro_export]
+macro_rules! inner_getters {
+    (self.$v:ident: $t:ty) => {
+        /// Gets access to the inner view.
+        pub fn get_inner(&self) -> &$t {
+            &self.view
```
hcpl: Should this be `&self.$v`? The same applies below.

xliiv: Thanks
https://github.com/gyscos/Cursive/pull/198

gyscos: Good catch! I should really pay more attention... <_<

PR #195, opened 2018-01-26
Change OnEventView.inner to OnEventView.view

xliiv:
This breaks compatibility!


gyscos:
Looks good!

Does it even break compatibility though? This was a private field.

xliiv:
You're right. I have to finally convert myself to rust.

PR #200, opened 2018-02-01
Moving Windows in StackView

IGBC:
This PR adds the ability to move a child of a stackView in the screen space using a new function `reposition_layer()`. With this functionality new handling of exposed background space was needed, so this responsibility was moved from `cursive::Cursive` into the stackView. Handling of background drawing was changed from a `clear()` call, which is unavailable to a series of printing commands, gated by a dirty flag. This new drawing method sees a performance improvement in applications previously making a lot of clear calls.

 Currently all commands that previously invoked `clear()` in `cursive::Cursive` now set the dirty flag:

- `pop_layer()`
- `reposition_layer()`
- handling `WindowChange` Events

Finally an example `./examples/position.rs` was crafted that demonstrates a movable window, and includes the menubar example to prove the new rendering does not break menu bars.

- **Rustfmt:** Formatted
- **Tests:** Passed Locally

_This is my first PR. Be gentle_

\- Segfault.

### DIFF
--- a/examples/position.rs
+++ b/examples/position.rs
@@ -1,36 +1,112 @@
 extern crate cursive;
 
 use cursive::Cursive;
-use cursive::views::TextView;
-use cursive::views::LayerPosition;
+use cursive::event::Key;
+use cursive::menu::MenuTree;
+use cursive::traits::*;
 use cursive::view::Position;
-
+use cursive::views::Dialog;
+use cursive::views::LayerPosition;
+use cursive::views::TextView;
+use std::sync::atomic::{AtomicUsize, Ordering};
 
 /// Moves top layer by the specifyed amount
-fn move_top(mut c: &mut Cursive, x_in: isize, y_in: isize) {
+fn move_top(c: &mut Cursive, x_in: isize, y_in: isize) {
     {
         // Step 1. Get the current position of the layer.
-        let mut s = c.screen_mut();
+        let s = c.screen_mut();
         let l = LayerPosition::FromFront(0);
-        let (x,y) = s.offset().pair();
+        let (x, y) = s.offset().pair();
 
         // Step 2. add the specifed amount
         // (unsigned math in Rust is a mess.)
-        let x = (if x_in < 0 { x - (-x_in) as usize } else { x + x_in as usize });
-        let y = (if y_in < 0 { y - (-y_in) as usize } else { y + y_in as usize });
-        
+        let x = if x_in < 0 {
+            x - (-x_in) as usize
+        } else {
+            x + x_in as usize
+        };
+        let y = if y_in < 0 {
+            y - (-y_in) as usize
+        } else {
+            y + y_in as usize
+        };
+
         // convert the new x and y into a position
-        let p = Position::absolute((x,y));
+        let p = Position::absolute((x, y));
 
         // Step 3. Apply the new position
         s.reposition_layer(l, p);
     }
     // Step 4. clean the screen cos we made it dirty.
-    c.clear();
+    //c.clear();
 }
 
 fn main() {
     let mut siv = Cursive::new();
+
+    // We'll use a counter to name new files.
+    let counter = AtomicUsize::new(1);
+
+    // The menubar is a list of (label, menu tree) pairs.
+    siv.menubar()
+        // We add a new "File" tree
+        .add_subtree("File",
+             MenuTree::new()
+                 // Trees are made of leaves, with are directly actionable...
+                 .leaf("New", move |s| {
+                     // Here we use the counter to add an entry
+                     // in the list of "Recent" items.
+                     let i = counter.fetch_add(1, Ordering::Relaxed);
+                     let filename = format!("New {}", i);
+                     s.menubar().find_subtree("File").unwrap()
+                                .find_subtree("Recent").unwrap()
+                                .insert_leaf(0, filename, |_| ());
+
+                     s.add_layer(Dialog::info("New file!"));
+                 })
+                 // ... and of sub-trees, which open up when selected.
+                 .subtree("Recent",
+                          // The `.with()` method can help when running loops
+                          // within builder patterns.
+                          MenuTree::new().with(|tree| {
+                              for i in 1..100 {
+                                  // We don't actually do anything here,
+                                  // but you could!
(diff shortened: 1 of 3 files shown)


src/views/stack_view.rs
```diff
@@ -15,6 +16,7 @@ pub struct StackView {
     // Store layers from back to front.
     layers: Vec<Child>,
     last_size: Vec2,
+    dirty: cell::Cell<bool>,
```
gyscos: We could use a little comment here on what `dirty` is for.

examples/position.rs
```diff
@@ -0,0 +1,129 @@
...
+}
+
+fn main() {
+    let mut siv = Cursive::new();
+    siv.set_fps(60);
+
+    // We'll use a counter to name new files.
+    let counter = AtomicUsize::new(1);
+
+    // This is a direct copy of the menubar example.
```
gyscos: Maybe we should keep this example minimal, and focused on moving a layer. I think a simple window that can be moved around should be enough.

src/cursive.rs
```diff
@@ -416,9 +416,14 @@ impl Cursive {
 
     /// Convenient method to remove a layer from the current screen.
     pub fn pop_layer(&mut self) -> Option<Box<AnyView>> {
-        let result = self.screen_mut().pop_layer();
-        self.clear();
-        result
+        self.screen_mut().pop_layer()
+    }
+
+    /// Convenient stub forwaring layer repositioning.
```
gyscos: Small typo here: `forwaring` -> `forwarding`

src/cursive.rs
```diff
@@ -460,6 +465,15 @@ impl Cursive {
         let printer =
             Printer::new(self.screen_size(), &self.theme, &self.backend);
 
+        let selected = self.menubar.receive_events();
+
+        // Print the screen before the menubar
+        // else the screen might draw over it and hide it
```
gyscos: Unfortunately this leads to some regressions.

* First, `printer` is now re-defined before drawing the menubar, which means that:
    * The menubar inherits the focus state, which means it cannot be "focused" anymore (the selected item has a blue background rather than a red one)
    * The menubar inherits the offset, which means that when `autohide` is turned off, the menubar is offset when it shouldn't (and is 1 line too low).

This problem is easy to fix by scoping the `printer` variable:

```
{
        let offset = if self.menubar.autohide { 0 } else { 1 };
        let printer = printer.offset((0, offset), !selected);
        let id = self.active_screen;
        self.screens[id].draw(&printer);
}
// `Printer` re-definition ends here

// Now draw menubar...
```

* The second, more serious issue is that we **do** want the menubar to be drawn *before* the stackview, so that possible menus or other views are drawn above the menubar:

Current `master` behavior, where the menu is drawn over the bar:
screenshot from 2018-02-02 10-28-02

Behavior from this branch, where the menu is drawn under the bar:
screenshot from 2018-02-02 10-27-50

The problem lies in where we define the background. It used to be managed by `Cursive` itself, so it could fit things like the menubar between the background and the `StackView`. If it's part of the `StackView`, then there's no way to fit the menubar under it. This means either we put the menubar on top (as this PR does), which doesn't look quite as nice, or we let `StackView` manage the menubar (so *it* can draw it between the background and the views).
I haven't thought about it too much yet, so it may be a good idea, but it adds more complexity to `StackView`, which may not be what we want.

So having `StackView` handle its background internally makes the menubar situation a bit more tricky.
I think an easy fix is to have a compromise, where `StackView` handles the background, but not as part of its `draw` method: instead, it would have an extra method, maybe `StackView::pre_draw`, where the dirty flag is checked and the screen cleared. This method would be called by `Cursive` *before* drawing the menubar; and the `StackView` would be drawn after the menubar as it is on master.

This means instead of putting the background as part of `StackView`'s implementation of the `View` trait, it's now a responsibility of the `StackView` directly, just like adding/poping layers.

I think this would give us the flexibility we want, with little change.

src/views/stack_view.rs
IGBC: no problem, I just need to figure out how to check out the PR

src/cursive.rs
IGBC: I think the correct solution here is to be honest with the stack-view about it's size and offset, so that will not try to draw over the menu-bar. This leaves the result that menu items cannot cover the top bar, but in my eyes that is also a positive.

gyscos: This is the behavior with auto-hide disabled, but the current one is also something we want to support (it's what some applications like elinks do).

IGBC: I think you missunderstand. What I am suggesting is that we check if the menubar is currently displayed and if so call the draw phase of the SV with an offset of (0,1) and a reduced Hsize

IGBC: if it is not in view we call it with the full size.

IGBC: the change in view size will result in the dirty flag re-triggering.

gyscos: The elinks behavior I was refering to was the popup being drawn over the menubar - this is what wouldn't be possible by simply offsetting the `StackView`.

IGBC: its possible they are using the menubar widget manually not the one bundled into cursive.

gyscos: Yes, elinks isn't using cursive (it's a C program much older than either rust of cursive), but it's still a behavior I want to support in the core library.

gyscos: I think a good solution might be:
* When autohide is enabled, add the menubar as a view on the `StackView`. It will show above most windows, but under pop-up menus.
    * This means create a new menubar from the menu tree every time we want to make it appear, and pop it off when we want to hide it.
* When autohide is off, just show the menubar above the `StackView` as we do currently.

This needs a better support from `StackView` for menubar/statusbars - it may even help with other issues like #201 .

IGBC:
Pushed a new commit to master addressing all issues we discussed above. for menubar drawing I built a compromise which separates background and foreground drawing into separate functions which are called either side of the menubar drawing, these are then merged back together in the draw() call (which is not used in the cursive::Cursive anymore) allowing for predictable behaviour if the stackview is nested (a behaviour my applications depend on)

gyscos:
Looks great, thanks again!
For the signed/unsigned mess, I'll push shortly a small improvement using `XY<isize>`.

PR #204, opened 2018-02-08
Add LinearLayout.get_focus_index

### LINKED ISSUE #203: [question] How to determine which view has focus
How to determine which view has focus, see `wrap_on_event` body for details 

```rust
#[macro_use]
extern crate cursive;
use std::any::Any;

use cursive::Cursive;
use cursive::Printer;
use cursive::traits::*;
use cursive::view::ViewWrapper;
use cursive::views::*;

struct Composed1 {
    view: LinearLayout,
}

impl Composed1 {
    fn new() -> Self {
        let mut layout = LinearLayout::vertical()
            .child(EditView::new())
            .child(EditView::new());

        Composed1 {
            view: layout,
        }
    }
}

impl ViewWrapper for Composed1 {
    wrap_impl!(self.view: LinearLayout);

    fn wrap_on_event(&mut self, event: Event) -> EventResult {
        eprintln!("{:?}", event);
        match event {
            Event::Key(Key::Enter) => {
                // how to determine which one of two EditView is focused?
                if first_edit_focused {
                    eprintln!("first focused");
                } else {
                    eprintln!("second focused");
                }
            }
        }
        EventResult::Consumed(None)
    }
}


fn main() {
    let mut siv = Cursive::new();

    let widget = Composed1::new();
    siv.add_layer(Dialog::around(widget).fixed_width(20));

    siv.run();
}
```

xliiv:
Fix #203

### DIFF
--- a/src/views/linear_layout.rs
+++ b/src/views/linear_layout.rs
@@ -145,6 +145,11 @@ impl LinearLayout {
         self.invalidate();
     }
 
+    /// Returns index of focused inner view
+    pub fn get_focus_index(&self) -> usize {
+       self.focus
+    }
+
     // Invalidate the view, to request a layout next time
     fn invalidate(&mut self) {
         self.cache = None;


src/views/linear_layout.rs
```diff
@@ -145,6 +145,11 @@ impl LinearLayout {
         self.invalidate();
     }
 
+    /// Returns index of focused inner view
+    pub fn get_focus_index(&self) -> &usize {
```
xliiv: or `usize`?

hcpl: For types with small size like primitives, it's preferable to return by value since it's often cheaper to copy them than to provide a pointer indirection.

Also, an immutable reference to `LinearLayout`'s field may interfere with the need to use mutable references to `LinearLayout` or to move it altogether according to borrowing rules. If that's what you want to keep the value synchronized with the state of a `LinearLayout` without having to write the synchronization code, then that's fine. In most other cases it's better to return `usize`.

xliiv: thanks for clarification.
Updated

gyscos:
Thank you very much for the PR!

PR #208, opened 2018-02-11
Allow FnOnce callbacks in cb_sink

ngg:
This is unfortunately changing the API but it makes dealing with callbacks much easier.
This PR also modifies that multiple callbacks can be run at each step. (The documentation suggested it worked like this before: "Callbacks will be executed in the order of arrival on the next event cycle.")

### DIFF
--- a/src/cursive.rs
+++ b/src/cursive.rs
@@ -15,6 +15,24 @@ use views::{self, LayerPosition};
 /// Identifies a screen in the cursive root.
 pub type ScreenId = usize;
 
+/// Asynchronous callback function trait.
+///
+/// Every `FnOnce(&mut Cursive) -> () + Send` automatically
+/// implements this.
+///
+/// This is a workaround only because `Box<FnOnce()>` is not
+/// working and `FnBox` is unstable.
+pub trait CbFunc: Send {
+    /// Calls the function.
+    fn call_box(self: Box<Self>, &mut Cursive);
+}
+
+impl<F: FnOnce(&mut Cursive) -> () + Send> CbFunc for F {
+    fn call_box(self: Box<Self>, siv: &mut Cursive) {
+        (*self)(siv)
+    }
+}
+
 /// Central part of the cursive library.
 ///
 /// It initializes ncurses on creation and cleans up on drop.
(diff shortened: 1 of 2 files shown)


hellow554:
How does it work in code? Can you provide an example/adapt the examples?

ngg:
Except for the type of `cb_sink` nothing is changed when using it, for example the only current example that uses it (progress.rs) doesn't need any change at all. The difference is that before this patch only Fn functions could be sent, now you can send functions that are FnMut or FnOnce only. For example if a closure needs mutable access to its captures or it needs to move them then it is not Fn, but still can be used.

src/cursive.rs
```diff
@@ -526,8 +542,8 @@ impl Cursive {
     ///
     /// [`run(&mut self)`]: #method.run
     pub fn step(&mut self) {
-        if let Ok(cb) = self.cb_source.try_recv() {
```
gyscos: No idea why I didn't use a `while` here - thanks!

src/cursive.rs
```diff
@@ -15,6 +15,22 @@ use views::{self, LayerPosition};
 /// Identifies a screen in the cursive root.
 pub type ScreenId = usize;
 
+/// Asynchronous callback function trait.
+///
+/// Every FnOnce(&mut Cursive) -> () + Send automatically
+/// implements this. This is a workaround only because
+/// Box<FnOnce()> is not working and FnBox is unstable.
+pub trait CbFunc: Send {
```
gyscos: Thanks! I think I got disapointed by the `FnBox` situation and just stopped thinking about it, without realizing a workaround like this could work. Awesome!

src/cursive.rs
```diff
@@ -81,7 +97,7 @@ impl Cursive {
     /// regularly check for messages.
     ///
     /// [`set_fps`]: #method.set_fps
-    pub fn cb_sink(&self) -> &mpsc::Sender<Box<Fn(&mut Cursive) + Send>> {
+    pub fn cb_sink(&self) -> &mpsc::Sender<Box<CbFunc>> {
```
gyscos: I think a small example code here, or a link to an example, would help users understand that they can just send regular closures here.

hellow554:
Okay,  thanks for the clarification 😃

ngg:
I've added an example with a closure to cb_sink() and fixed the docs. Please make sure to bump to 0.8 if you will merge this (API change).

gyscos:
Perfect! I was going to bump the version anyway, so perfect timing.
I just realized we now need to explicitly specify the type in the closure given, but that's fine.

ngg:
Yes, that's unfortunate but probably not too bad. I'm not sure why the compiler cannot infer that, they might fix it someday :)

hellow554:
@ngg This patch made poeple at stackoverflow happy :) 
https://stackoverflow.com/questions/49142453/is-it-possible-to-call-a-fnonce-from-a-fn-without-a-mutex/49148061#49148061
Maybe you want to publish it as a blog post or similar? This seems to be a common problem right now.

PR #211, opened 2018-02-18
Add get_current_focus to Dialog & make public enum Focus

xliiv:


src/views/dialog.rs
```diff
@@ -14,8 +14,13 @@ use view::{AnyView, Selector, View};
 use views::{Button, DummyView, SizedView, TextView};
 
 #[derive(PartialEq)]
-enum Focus {
+/// Identifies currently focused elements in [`Dialog`].
+///
+/// [`Dialog`]: struct.Dialog.html
+pub enum Focus {
```
gyscos: I think we should rename this to `DialogFocus`, otherwise, it's not clear from `cursive::views::Focus` what this refers to.

src/views/dialog.rs
```diff
@@ -244,6 +249,11 @@ impl Dialog {
         Box::new(self.buttons.iter_mut().map(|b| &mut b.button.view))
     }
 
+    /// Returns currently focused element
+    pub fn get_current_focus(&self) -> &Focus {
```
gyscos: Let's rename that to simply `focus(&self)` - there can be no other (non-current) focus, and most getters don't need a `get_` prefix.

src/views/dialog.rs
xliiv: It was one of mine concerns.. . I'll change it.

xliiv:
I've changed it, take a look if it's ok now @gyscos

src/views/dialog.rs
```diff
@@ -14,8 +14,13 @@ use view::{AnyView, Selector, View};
 use views::{Button, DummyView, SizedView, TextView};
 
 #[derive(PartialEq)]
-enum Focus {
+/// Identifies currently focused element in [`Dialog`].
+///
+/// [`Dialog`]: struct.Dialog.html
+pub enum DialogFocus {
```
gyscos: Let's add the usual `#[derive(Clone,Copy,Debug,PartialEq,Eq)]`, since it's now a public field.

src/views/dialog.rs
```diff
@@ -244,6 +249,11 @@ impl Dialog {
         Box::new(self.buttons.iter_mut().map(|b| &mut b.button.view))
     }
 
+    /// Returns currently focused element
+    pub fn focus(&self) -> &DialogFocus {
```
gyscos: Since `DialogFocus` is so small (it's barely lareger than a `usize`), and it'll be `Copy`, let's directly return a `DialogFocus` here without references. It'll prevent borrowchecker confusions later on... :)

src/views/dialog.rs
xliiv: done

src/views/dialog.rs
xliiv: done

xliiv:
Fixed :)

gyscos:
Thanks again!

xliiv:
My pleasure ;P

PR #213, opened 2018-03-02
Fix StackView.move_layer

xliiv:


gyscos:
Ah yes, I see the problem.

I see two main ways to translate the user directive into an actual placement. (Note: the current solution is neither, and should be fixed no matter what).

* The given destination indicates where to insert the view.
    Example:
    schema
    
    This is what I was trying to achieve, but it's currently broken for `FromBack` indices.
    The confusing part is that the destination value is not always the actual position the view will have after moving: for example, moving `FromFront(1)` to `FromFront(3)` will put the view 1 just before view 3, which means its actual position will end up being `FromFront(2)`.

* The given destination represents the position the layer will be.
  This eliminates the confusing point above: after moving any view to `FromFront(3)`, we know its position will actually be `FromFront(3)`.
  The confusing part is that when moving a view to `FromFront(17)`, we don't know if the view will end up being *before* or *after* the current "View 17": this depends on whether the view we are moving is *currently* higher or lower than 17. So if I'm given `n` and `m` and want the view `n` to be just in front of view `m`, I will need to move it to `FromFront(m)` if `n > m`, and `FromFront(m+1)` if `n < m`.

I guess both solutions are confusing in their own ways, but I suppose the second one is less so.

xliiv:
I don't understand this explanation or I don't see the problem.
The one one problem i'd see is the mutation of array, but once indexes are counted before mutation, everything should works fine.
Could you write simple a test case then i add it to code and then maybe i'll see the problem?

gyscos:
It's not a technical problem, it's just 2 possible interpretations of the destination, each with its surprising part.

That being said, I agree the second solution is probably the best overall; I'll fix that soon.

xliiv:
Ok, then i'm closing it.

PR #215, opened 2018-03-11
Option to include identifier in stackview and move by id

yeastplume:
Hello, I'm trying to integrate Cursive into the Grin Cryptocurrency project (https://github.com/mimblewimble/grin), and am having difficulty trying to create a simple always-displayed menu that changes the content of another view onscreen. The issue is that there's no way to identify a stack view other than by position in the stack, which makes it cumbersome to relate a menu item to a particular stack view. 

I've forked off my own branch to include in the project, but it would be very good if this, (or something equivalent) could be included upstream. Basically, it changes the 'layer' function in StackView to include an optional id:

```
let stack = StackView::new()
   .layer("TextView::new("1"), Some("layer_1"))
   .layer("TextView::new("2"), Some("layer_2"))
   .layer("TextView::new("1"), Some("layer_3"))
```

And then layers can be brought to the front with:

```
   stack.move_id_to_front("layer_2");
```

etc.

Not offended if what I have here breaks any design principles, but I think this feature would be very good to include in some form or another. I'd originally tried to only include the ID in the Child struct based on whether the added view was wrapped in an IDView, but had trouble trying to downcast it generically to operate on it (I don't think it can be done in Rust), so unfortunately my approach required an additional Option argument to `layer` and related functions.

Thank you for your work so far on Cursive, it seems excellent.

### DIFF
--- a/examples/mutation.rs
+++ b/examples/mutation.rs
@@ -41,6 +41,7 @@ fn show_popup(siv: &mut Cursive) {
                 });
             })
             .dismiss_button("Ok"),
+        None,
     );
 }
 
--- a/src/cursive.rs
+++ b/src/cursive.rs
@@ -455,7 +455,7 @@ impl Cursive {
     /// # }
     /// ```
     pub fn add_layer<T: 'static + View>(&mut self, view: T) {
-        self.screen_mut().add_layer(view);
+        self.screen_mut().add_layer(view, None);
     }
 
     /// Adds a new full-screen layer to the current screen.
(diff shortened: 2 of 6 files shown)


gyscos:

Indeed, it is currently quite inconvenient to get a view position from its ID.

06305c89a9223ffa0b041c94df4a51a177b1c99a should bring `StackView::find_view_from_id`, which allows you to identify the layer containing an ID (the view with the ID could be deeply nested within the layer).

This should let you find the position of the view you want to move, without affecting users who don't need this.

gyscos:
I'm glad I could help! Don't hesitate to ask for anything, or to report any painful point when using cursive - feedback is golden!
Good luck to you guys, grin looks like a pretty nice project to follow! :)

PR #228, opened 2018-03-28
Adding disable/enable functionality to text_area

itgonzo:
Adding disable/enable functionality that seems to be missing in TextArea


gyscos:
Thanks for the PR! Not sure why this wasn't the before... :S

PR #229, opened 2018-04-01
Making `backend::Backend` into a `std::Box`able trait 

IGBC:
# Summary
[summary]: #summary

This PR modifies the `Backend` trait and its accompanying `Concrete` implementations to support boxing the trait into a trait object, and modifying the `Cursive` and `Printer` objects to use this boxed trait.

# Motivation
[motivation]: #motivation

The existing `Printer` and `Cursive` implementations are hardcoded to use the `backend::Concrete` backend implementation. This causes no issues for the users of the Cursive library, but 3rd party developers (like me) wishing to embed Cursive into other frameworks cannot replace the Concrete implementation with their own backend.

This PR allows two major changes for 3rd party developers:
  * They can create structs that contain `Box<backend::Backend>` fields that can be populated with a backend at runtime.
  * Custom Backends can be passed to `Printer` allowing Cursive to be used with custom backends.

# Drawbacks
[drawbacks]: #drawbacks

  * Instantiating printers is now slightly more complicated. 
  * Breaking Change to backend API - All implementors have been updated.
  * Backend trait is now usable outside of library, so should be semantically versioned, from here out.

# Side-effects 

- Reverse Video Effect in BLT has been fixed and now supports color

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -35,8 +35,10 @@ pub trait Backend {
 
     fn set_refresh_rate(&mut self, fps: u32);
 
+    // This sets the Colours and returns the previous colours
+    // to allow you to set them back when you're done.
     fn set_color(&self, colors: theme::ColorPair) -> theme::ColorPair;
 
     fn set_effect(&self, effect: theme::Effect);
-    fn reset_effect(&self, effect: theme::Effect);
+    fn unset_effect(&self, effect: theme::Effect);
 }
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -16,7 +16,7 @@ pub use self::curses::*;
 pub use self::termion::*;
 
 pub trait Backend {
-    fn init() -> Self;
+    fn init() -> Box<Self> where Self: Sized;
     // TODO: take `self` by value?
     // Or implement Drop?
     fn finish(&mut self);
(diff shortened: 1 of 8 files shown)


gyscos:
Thank you very much for the work!

Making the backend dynamic has been a goal for a while, this will definitely help many use-cases!

src/backend/mod.rs
```diff
@@ -34,7 +34,9 @@ pub trait Backend {
     fn clear(&self, color: theme::Color);
 
     fn set_refresh_rate(&mut self, fps: u32);
-    // TODO: unify those into a single method?
-    fn with_color<F: FnOnce()>(&self, colors: theme::ColorPair, f: F);
-    fn with_effect<F: FnOnce()>(&self, effect: theme::Effect, f: F);
+
+    fn set_color(&self, colors: theme::ColorPair) -> theme::ColorPair;
```
gyscos: Can we add a comment here to indicate that the backend should return the previous state?

src/backend/mod.rs
```diff
@@ -34,7 +34,9 @@ pub trait Backend {
...
+
+    fn set_effect(&self, effect: theme::Effect);
+    fn reset_effect(&self, effect: theme::Effect);
```
gyscos: Can we call it `unset_effect`? `reset` can be a bit ambiguous sometimes.

IGBC:
yea we can do both of those, actually in one of the implementations I actually typoed one of the implementations to "unset" XD

IGBC:
I have updated as requested

src/printer.rs
```diff
@@ -165,7 +166,9 @@ impl<'a> Printer<'a> {
     where
         F: FnOnce(&Printer),
     {
-        self.backend.with_effect(effect, || f(self));
+        self.backend.set_effect(effect);
+        f(self);
+        self.backend.reset_effect(effect);
```
gyscos: Just this one left to update

PR #246, opened 2018-05-16
Attempted to create a side channel to inject artificial events.

njskalski:
Hi. I attempted to create a side channel to inject artificial events. The use case is a refresh between the inputs. You may consider the approach dirty, but that was best I could come up to between midnight and 2am.

I should probably add some tests, cleanup warnings and write some docs, but before I move forward I would like to hear what do you think.

This is a "workaround", not proper solution, but I'd really use it. I'm basically building a command-line IDE for modern languages (Rust, Dart, Go), and being able to refresh after a long running task finishes (indexing, kythe.io lookup, LSP lookup, syntax coloring finished, any plugin non-blocking interaction) is crucial for UX.

### DIFF
--- a/examples/lorem.rs
+++ b/examples/lorem.rs
@@ -1,9 +1,8 @@
 extern crate cursive;
 
-use cursive::Cursive;
 use cursive::align::HAlign;
-use cursive::view::Boxable;
 use cursive::views::{Dialog, Panel, TextView};
+use cursive::Cursive;
 
 fn main() {
     // Read some long text from a file.
@@ -18,11 +17,10 @@ fn main() {
     // and will adapt to the terminal size.
     siv.add_fullscreen_layer(
         Dialog::around(Panel::new(TextView::new(content)))
+            .title("Unicode and wide-character support")
             // This is the alignment for the button
             .h_align(HAlign::Center)
-            .button("Quit", |s| s.quit())
-            .title("Unicode and wide-character support")
-            .full_screen(),
+            .button("Quit", |s| s.quit()),
     );
     // Show a popup on top of the view.
     siv.add_layer(Dialog::info(


njskalski:
yes, the question is whether it is desired to expose such option. Imagine you do, and you have to support it later. That would affect refactoring options in the future. I clearly see a use case for refresh, did not really think about other events.

njskalski:
@zpgaal but if that was review and you want to get it exposed I can do it. @gyscos what do you think?

zpgaal:
It was just a quick look at the API change. I leave the review to the repo owner :).

njskalski:
Hi, I do not understand what your solution would look like. If I did, maybe I'd code it myself to speed things up. For purpose of my application set_fps is not really acceptable, especially with termion it leads to screen flickering even on strong machines, and my program is targeted rather towards older (say 4-5 years) machines.

I am still waiting for clearance to share my code.

gyscos:
I almost have a prototype ready, but I'm having issues with the BearLibTerminal backend; it seems it's really not thread-safe, and crashes with `[fatal] 'read' was not called from the main thread` if `wait_event` is called from a separate thread... :confused: 

As for the other backends, we can't really "make sure that ncurses is never accessed by two threads at the same time", since this is the whole point: we want to be able to refresh the screen (write to ncurses) while we are still waiting for an event (`ncurses::getch()` is being called).

njskalski:
Only one thing comes to my mind: signals. I think we can signal SIGINT from another thread and upon receiving Err see if that signal is expected (was triggered by us) or was actually sent from outside the program. I know it's a hack, but that's only thing that comes to my mind.

https://docs.rs/nix/0.10.0/nix/sys/signal/fn.raise.html

gyscos:
If the lib is not thread-safe, it means it shares some internal variables between methods - I don't think interrupting these methods with a signal is really safe. Plus, BearLib works on windows, where I have no idea how signals work.

I think for BearLibTerminal I'll either not support interruptions, or poll the event queue every 30ms or so (basically forcing a 30FPS refresh rate). We'll see later if we find smarter solutions for this backend.

PR #251, opened 2018-05-21
added with_theme to Printer

njskalski:


njskalski:
cool. Can you push it to 0.8.1 crate?

njskalski:
To me it is. That would be appreciated. Thanks.

gyscos:
Well, 0.8.2 is out, that's pretty easy.
For 0.9, the `scroll` branch is making good progress, and so is `callback_preemption`. Shouldn't be _much_ longer... :^/

Issue #117, opened 2017-03-15
Padding for LinearLayout


marshauf:
Hello,

I would like to add a row between a PogressBar and a Button which are in a LinearLayout.
I couldn't find a way to separate the two Views from each other. I would prefer to set a value like in a Dialog with padding.
Can you add one or do I need to add a DummyView between the two to achieve my goal?

gyscos:
Dummy views are there for this purpose.

Though I can understand it'd start to be annoying having to add one every time if you have a very long LinearView... 

I'll look into that. Maybe it could also have optional borders...

dermetfan:
**TL;DR**: How do I build a transparent `DummyView`?

I'm trying to visualize the keymap in my editor built on Cursive.

Currently I position the keymap view below the editor using a `LinearLayout`:

```rust
LinearLayout::vertical()
    .child(view.with_id("inner")            .full_screen()) // main editor
    .child(Line::horizontal()               .full_width())  // delimiter
    .child(ListView::new().with_id("keymap").full_width())  // keymap
    .full_width()
    .with_id("overlay")
```

Resulting in (list is empty):

![2018-05-20t14 21 30 435431908 02 00](https://user-images.githubusercontent.com/4956158/40278893-acaaa682-5c39-11e8-9b9c-533729c9bbe1.png)

However, since the cursor is always scrolled into view, the text scrolls when I fill the keymap view and the view size changes:

![2018-05-20t14 21 39 050718751 02 00](https://user-images.githubusercontent.com/4956158/40278894-accd7f0e-5c39-11e8-947f-fbce6e0a0e18.png)

I would like the keymap to behave like an overlay though, with no influence on the content behind it. The delimiter should also only be shown when the overlay is shown. So I wrapped everything in a `StackView` from which I can just pop off the overlay. To keep it at the bottom of the screen though, I need to add a `DummyView` that acts as spacer.

However, even though `DummyView` does nothing in `draw()`, the editor is not visible behind it:

![2018-05-20t14 22 18 541359216 02 00](https://user-images.githubusercontent.com/4956158/40278895-ace97e20-5c39-11e8-86eb-551c58cb23ca.png)

Do you have any tips on how I could achieve that layout?

gyscos:
Hi!

`DummyView` is transparent, the problem is the layer itself. `StackView` wraps each child in a `Layer` which fills the background for this view, which is why layers are not transparent.

[...]

PR #254, opened 2018-05-31
transparent StackView layers

dermetfan:
In response to https://github.com/gyscos/Cursive/issues/117#issuecomment-390495759 I removed the `Layer` from `StackView`, added a background to `Dialog` and `Panel`, and adapted usages. On the caller's side that leads to the following situation:

`add_layer(Layer::new(…))`

By default there will be no background so in its current state this silently breaks many applications and is not very intuitive / easy to do wrong. You can see how many occurrences I had to update in the examples. Because the background is not cleared this also causes artifacts when scrolling text.

Maybe we should keep the default behavior and just add something like `add_transparent_layer()`? I'd like to agree on the API changes before I go ahead.

### DIFF
--- a/src/views/stack_view.rs
+++ b/src/views/stack_view.rs
@@ -57,8 +57,10 @@ impl Placement {
 enum ChildWrapper<T: View> {
     // Some views include a shadow around.
     Shadow(ShadowView<Layer<T>>),
+    // Some include a background.
+    Backfilled(Layer<T>),
     // Some views don't (fullscreen views mostly)
-    Plain(Layer<T>),
+    Plain(T),
 }
 
 impl<T: View> ChildWrapper<T> {
@@ -68,7 +70,11 @@ impl<T: View> ChildWrapper<T> {
             ChildWrapper::Shadow(shadow) => {
                 shadow.into_inner().ok().unwrap().into_inner().ok().unwrap()
             }
-            ChildWrapper::Plain(layer) => layer.into_inner().ok().unwrap(),
+            // Layer::into_inner can never fail.
+            ChildWrapper::Backfilled(background) => {
+                background.into_inner().ok().unwrap()
+            }
+            ChildWrapper::Plain(layer) => layer,
         }
     }
 }
@@ -78,7 +84,8 @@ impl<T: View> ChildWrapper<T> {
     pub fn get_inner(&self) -> &View {
         match *self {
             ChildWrapper::Shadow(ref shadow) => shadow.get_inner().get_inner(),
-            ChildWrapper::Plain(ref layer) => layer.get_inner(),
+            ChildWrapper::Backfilled(ref background) => background.get_inner(),
+            ChildWrapper::Plain(ref layer) => layer,
         }
     }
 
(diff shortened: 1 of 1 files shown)


gyscos:
Thank you very much for the work!

You're right, silently changing `add_layer` to be transparent might be surprising; a `add_transparent_layer` sounds like a better solution.

A simple solution would be, from your PR, to simply rename `add_layer` to `add_layer_transparent`, and add something like

```rust
fn add_layer<T>(&mut self, view: T)
    where T: IntoBoxedView,
{
    self.add_transparent_layer(Layer::new(view));
}
```

Unfortunately, thinking about it more, we have the same problem we had with the optional `ShadowView` wrapper: we want users who pop layers to get the `View` they inserted. So if I do `add_layer(MyView::new())`, I want `pop_layer` to return this `MyView`, without being wrapped by a `Layer` or `ShadowView`.

The solution is the same we had back then: add a variant to the `ChildWrapper` enum, so that it could be either `Shadow(ShadowView<Layer<T>>)`, `Backfilled(Layer<T>)`, or `Transparent(T)`.

This means adding `add_transparent_layer` and `add_transparent_layer_at`.

This may sound like an uncontrolled explosion in methods or variants, but:
* The variants for `ChildWrapper` are bounded: we did not _add_ a new wrapper, we _removed_ the last mandatory wrapper.
* We could remove `add_layer_at` and only keep `add_transparent_layer_at`: it's an advanced API to give more control; we can expect users in this case to be fine with manually wrapping a view in a `Layer`.

Going this road, I don't think it makes sense to backfill in Dialogs or Panels anymore.

dermetfan:
That turned out pretty elegant, only `StackView` is modified now. Thanks for the input!
I also added `transparent_layer()` for a chainable variant.

src/views/stack_view.rs
```diff
@@ -57,8 +57,10 @@ impl Placement {
 enum ChildWrapper<T: View> {
     // Some views include a shadow around.
     Shadow(ShadowView<Layer<T>>),
+    // Some include a background.
+    Backfilled(Layer<T>),
     // Some views don't (fullscreen views mostly)
-    Plain(Layer<T>),
+    Plain(T),
```
dermetfan: (Kept the `Plain` name for now, I think it's still a good fit)

src/views/stack_view.rs
```diff
@@ -77,18 +83,22 @@ impl<T: View> ChildWrapper<T> {
     /// Returns a reference to the inner view
     pub fn get_inner(&self) -> &View {
         match *self {
-            ChildWrapper::Shadow(ref shadow) => shadow.get_inner().get_inner(),
-            ChildWrapper::Plain(ref layer) => layer.get_inner(),
+            ChildWrapper::Shadow(ref shadow) => shadow.get_inner(),
```
gyscos: I think we still want the absolute inner view here, so it would be `get_inner().get_inner()` (just like before this PR)?

src/views/stack_view.rs
```diff
@@ -77,18 +83,22 @@ impl<T: View> ChildWrapper<T> {
...
+            ChildWrapper::Plain(ref layer) => layer,
         }
     }
 
     /// Returns a mutable reference to the inner view
     pub fn get_inner_mut(&mut self) -> &mut View {
         match *self {
             ChildWrapper::Shadow(ref mut shadow) => {
-                shadow.get_inner_mut().get_inner_mut()
+                shadow.get_inner_mut()
```
gyscos: Same thing here, any reason you removed a round of `get_inner_mut()`?

dermetfan:
Good catch! I added a test that would have caught my mistake.

src/views/stack_view.rs
dermetfan: Those are my fault, it seems I haven't properly restored everything after the previous attempt.

gyscos:
Awesome, thanks!

PR #255, opened 2018-06-05
Don't require num default features

sportzer:
The num crate's default features pull in some extra stuff which isn't used by Cursive. Most notably, rustc-serialize, which prevents compiling to wasm.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,13 +17,16 @@ repository = "gyscos/Cursive"
 enum-map = "0.2.24"
 enumset = "0.3.3"
 log = "0.4"
-num = "0.1"
 owning_ref = "0.3.3"
 toml = "0.4"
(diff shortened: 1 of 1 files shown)


gyscos:
Interesting, I didn't think of checking that.

Thanks for the PR!

hellow554:
Cursive in wasm? 😱 I want to see that! Sounds amazing!

sportzer:
It's a super hacky prototype, but I'm basically compiling to wasm32-unknown-unknown and using the custom backend support that's coming in Cursive 0.9 to expose terminal state to javascript. The actual rendering uses [rot.js](http://ondras.github.io/rot.js/hp/). My code is [here](https://github.com/sportzer/cyberphage), and it's semi-playable over on [itch.io](https://sportzer.itch.io/cyberphage).

PR #256, opened 2018-06-08
changed the Palette type and introduced Index trait implementations

njskalski:


gyscos:
I've been thinking more about what the API would look like, and I ended up with something like this:

```rust
// Let's use the defaut palette for this example.
let palette = Palette::default();

// Implement Index<PaletteColor> just like in this PR
let color: Color = palette[PaletteColor.Primary];

// Return an `Option` on custom key access
let color: Option<Color> = palette.custom("MY CUSTOM KEY");

// Merge a namespace in the top-level palette
let sub_palette: Palette = palette.merge("MY NAMESPACE");
```

For custom colors, I prefer to return an `Option` than to specify the fallback directly in the getter - this allows more flexible usage, for instance using another custom key as fallback. But the `Index` trait doesn't sound ideal to return an `Option` (the standard library would have it panic instead, which is not really desirable here), so a getter method (here `fn custom<'a>(&self, key: &'a str) -> Option<Color>`) sounds like a better fit.

Finally, merging namespaces means there should be little no need for nested getters: instead of `palette["namespace"]["key"]` or `palette["namespace/key"]`, it would be `palette.merge("namespace").custom("key")`.

gyscos:
I implemented something like that in 017e100b1b4be67b1f3df55e89e3b675a4438a42: the palette should read arbitrary values from the toml config, which can be accessed either through `Palette::custom()` or `Palette::merge`. In general, namespaces can be used to override values, and custom colors can be used by views that need more than the basic 10 colors.

njskalski:
that's cool. So now my offer is that I create a common repository of third party views (starting with the abandoned ones) and update them there. What do you think?

What version are we targeting these changes to?

gyscos:
All these changes should arrive in 0.9.0, along with the new `ScrollView` which just landed, and probably the callback preemption, which only needs a few fixes.

A common repository for third-party views sounds like a great idea, and could ease the maintenance for developers who can contribute some views but do not have time to maintain them. I don't think _all_ third-party views would need to be in a single repository, but at least for abandonned ones, it could act as a kind of nursery.

gyscos:
Closed by mistake.

Though some changes to the style system may be on the horizon.

PR #261, opened 2018-06-16
implement scroll_to_top and scroll_to_bottom

sajattack:


gyscos:
Thanks a lot for this work!

A bit of nitpicking: `ScrollView` can scroll horizontally as well as vertically.
Therefore, maybe we should name it `scroll_to_top_left` and `scroll_to_bottom_left`?

sajattack:
Or should we keep the X value rather than set it to 0?

PR #272, opened 2018-07-25
Fix a few typos in tutorial 1

robbystk:
\<Ctrl-C\> gets interpreted as an HTML tag as is so we need to put it in backtics.
An alternative is to escape the angle brackets with backslashes (\\\<Ctrl-C\\\>), but I think this doesn't look as good.

### DIFF
--- a/doc/tutorial_1.md
+++ b/doc/tutorial_1.md
@@ -89,7 +89,7 @@ fn main() {
 It's rather easy to identify the two steps involved.
 
 If you run this, you'll get an empty blue terminal, with no way of properly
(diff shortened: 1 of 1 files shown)


gyscos:
Whoops, thanks!

Another solution could be to use github's fancy `<kbd>` tag: <kbd>Ctrl</kbd>+<kbd>C</kbd>.

robbystk:
I didn't know about that, good to know.  I like that best of all.

gyscos:
Thanks again!

PR #275, opened 2018-08-01
Fix broken example in readme to use Cursive::default()

vlttnv:
Hi.
I just found out about this library and the first thing I did was to read the readme, create a new project and try the example. I was met with this error:
```rust
error[E0061]: this function takes 1 parameter but 0 parameters were supplied
 --> src/main.rs:8:19
  |
8 |     let mut siv = Cursive::new();
  |                   ^^^^^^^^^^^^^^ expected 1 parameter
```
I read the docs and switched the example to use `Cursive::ncurses()` instead. This should help new users like me and provide them with a smooth introduction into the library.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -37,7 +37,7 @@ use cursive::views::{Dialog, TextView};
 
 fn main() {
     // Creates the cursive root - required for every application.
-    let mut siv = Cursive::ncurses();
+    let mut siv = Cursive::default();
 
     // Creates a dialog with a single "Quit" button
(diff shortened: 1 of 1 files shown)


gyscos:
Whoops, I have no excuse - thank you so much for noticing!
The recommended way is now `Cursive::default()`: it will adapt to whatever backend is compiled in.

vlttnv:
Haha, no worries. Just came here to say that `default` might be a better idea since I just noticed it's being used in the tutorial. Will update the pull request.

gyscos:
Thanks again!

Issue #278, opened 2018-08-07
Colored `ProgressBar`


HenryTheCat:
This is a feature request proposing an addition to the `ProgressBar` view.

# Problem description

Currently, `ProgressBar` uses the theme's `highlight` color as the bar's color, and that is not customizable. It may be desiderable to have progress bars with customized bar colors.

# Proposal

Extend the `ProgressBar` API in a backward-compatible way to optionally specify a custom color.

As a tentative implementation, a `custom_color: Option<Color>` field could be added to `ProgressBar`, initialized as `None` by default. New methods could be added to modify the new field. Then, when printing, `ProgressBar` would use its custom color if specified, or the theme's `highlight` otherwise.

HenryTheCat:
The issue looks simple enough that I may try implementing it myself, if the proposal is approved.

gyscos:
Hi, and thanks for the idea!

Adding such a custom color could be an option (though you may want to take a `ColorType` to let people use palette colors if they need).

Another option is something we've been working toward recently, which is locally overriding the theme. The idea is to derive a new theme just for this `ProgressBar`, by taking the parent theme and replacing the `Highlight` color from the palette with a new custom color. This effectively achieves the same result, but is a bit more general as it lets users override any color for any view.

The main issue with this solution is that users need to know _what_ color to override: they would need to read the documentation, see the mention of `Highlight`, and override this color. This is all a bit less obvious than an inherent method.

Note that these two methods are not incompatible - we may add a `custom_color` to `ProgressBar`, and still let users apply local themes overrides. This may be a way to see how convenient it is in real conditions.

TLDR: let's do it!

PR #279, opened 2018-08-07
Add custom color to `ProgressBar`

HenryTheCat:
This PR:

* Adds `color` component to `ProgressBar`.
* Modifies `ProgressBar::new` to default to `highlight` color.
* Modifies existing methods to use `color` component when printing.
* Adds relevant methods to change `color` component.
* Adds some comments to the relevant sections.

as per issue #278.

### DIFF
--- a/src/views/progress_bar.rs
+++ b/src/views/progress_bar.rs
@@ -1,7 +1,7 @@
 use align::HAlign;
 use std::cmp;
 use std::thread;
-use theme::{ColorStyle, Effect};
+use theme::{ColorStyle, ColorType, Effect};
 use utils::Counter;
 use view::View;
 use Printer;
@@ -16,6 +16,9 @@ use Printer;
 ///
 /// It also prints a customizable text in the center of the bar, which
 /// defaults to the progression percentage.
+///
+/// The bar defaults to the current theme's highlight color,
+/// but that can be customized.
 ///
 /// # Example
 ///
@@ -35,6 +38,7 @@ pub struct ProgressBar {
     min: usize,
     max: usize,
     value: Counter,
+    color: ColorType,
     // TODO: use a Promise instead?
     label_maker: Box<Fn(usize, (usize, usize)) -> String>,
 }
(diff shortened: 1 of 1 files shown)


src/views/progress_bar.rs
```diff
@@ -183,6 +188,26 @@ impl ProgressBar {
...
+    /// Sets the color style.
+    ///
+    /// Chainable variant of `set_color`.
+    pub fn with_color<C>(mut self, color: C) -> Self
+    where
+        C: Into<ColorType>,
+    {
+        self.color = color.into();
+
+        self
```
gyscos: A common trick for these methods is to use the `With` trait:

```rust
pub fn with_color<C>(mut self, color: C) -> Self
where
    C: Into<ColorType>,
{
    self.with(|s| s.set_color(color))
}
```
This way we don't need to copy code between the two methods.

HenryTheCat:
I am having second thoughts on this.

So, before the bar used the `highlight` color by picking it at every paint. May I assume that, when changing the `highlight` color, the bar would automatically adjust to the new one too?

Because now the `highlight` color is read and stored at the moment of creating the `ProgressBar` and from then on it won't change by itself.

To recover the original behaviour, we should rather wrap the bar's color in an option, so that the default `None` means whatever color `highlight` is at the moment of painting.

gyscos:
`ColorStyle::highlight()` returns a fixed pair of `ColorType`: `ColorType::Palette(PaletteColor::View)` for the foreground, and `ColorType::Palette(PaletteColor::Highlight)` for the background. These two values never change. This means `ColorStyle::highlight().bg` is always equal to `ColorType::Palette(PaletteColor::Highlight)`.

When used in `View::draw`, these `ColorType` values tell the printer to go look in its `Palette` for the actual color to use.

This means the palette lookup is done on every call to `draw` - so it's safe to set the `ColorType` when creating `ProgressBar`.

src/views/progress_bar.rs
HenryTheCat: I didn't know about the `With` trait. Thanks for pointing it out!

HenryTheCat:
That's great! I think I have no further objections.

PR #288, opened 2018-10-02
Add title to views::Panel

### LINKED ISSUE #216: Add title to views::Panel
Panels should really be able to have titles to behave as visual "view groups".

The last thing Dialog have is buttons, but I don't think it makes sense to add buttons to Panels.

jacobbudin:
This fixes #216.

It adds title and title position to `views::Panel`.

Example title in views::Panel

One question: I notice [`Dialog::set_title_position`](https://github.com/gyscos/Cursive/blob/92f919978c072540b41938ae1d3af547f010d58d/src/views/dialog.rs#L245-L249) does not invalidate. Is this intentional? I used this view's logic as the basis for my PR.

### DIFF
--- a/src/views/panel.rs
+++ b/src/views/panel.rs
@@ -1,22 +1,91 @@
+use align::*;
 use event::{Event, EventResult};
 use rect::Rect;
+use theme::ColorStyle;
+use unicode_width::UnicodeWidthStr;
 use vec::Vec2;
 use view::{View, ViewWrapper};
 use Printer;
+use With;
 
 /// Draws a border around a wrapped view.
 #[derive(Debug)]
 pub struct Panel<V: View> {
+    // Inner view
     view: V,
+
+    // Possibly empty title.
+    title: String,
+
+    // Where to put the title position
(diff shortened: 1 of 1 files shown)


src/views/panel.rs
```diff
@@ -47,4 +118,9 @@ impl<V: View> ViewWrapper for Panel<V> {
         let inner_size = size.saturating_sub((2, 2));
         self.view.important_area(inner_size) + (1, 1)
     }
+
+    fn wrap_needs_relayout(&self) -> bool {
+        self.invalidated
+            || self.with_view(|v| v.needs_relayout()).unwrap_or(true)
```
gyscos: In this case you can directly call `self.v.needs_relayout()` I think? The `with_view` interface is to deal with more complex wrappers where it can't always access the inner view (think `RefCell`).

jacobbudin:
@gyscos Thanks for the advice, help. I've made the correction to `wrap_needs_relayout`.

gyscos:
Thanks again! :)

PR #299, opened 2018-11-30
Add on_change callback

agavrilov:


src/views/radio.rs
```diff
@@ -120,15 +131,25 @@ impl<T> RadioButton<T> {
...
+        }
     }
 
     /// Selects this button, un-selecting any other in the same group.
     ///
     /// Chainable variant.
     pub fn selected(self) -> Self {
-        self.with(Self::select)
+        self.with(|s| {
+            s.select();
```
gyscos: Can you add a comment indicating that we're dropping the potential callback here on purpose?

src/views/radio.rs
```diff
@@ -70,6 +74,13 @@ impl<T> RadioGroup<T> {
     pub fn selection(&self) -> Rc<T> {
         self.state.borrow().selection()
     }
+
+    /// Sets a callback to be used when the selection changes.
+    pub fn set_on_change<F: 'static + Fn(&mut Cursive, &T)>(
```
gyscos: Can you add a chainable `pub fn on_change<F>(self, on_change: F) -> Self` as well?

agavrilov: Done

gyscos:
Thanks again!

PR #304, opened 2018-12-30
Hide Scrollbars

danielakhterov:


gyscos:
Hi, and thanks for the PR!

Whoops, it seems indeed that I didn't add methods to toggle this.

PR #306, opened 2019-01-07
Test backend

njskalski:
This is my early proposal for PuppetBackend

Please comment.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ license = "MIT"
 name = "cursive"
 readme = "Readme.md"
 repository = "https://github.com/gyscos/Cursive"
-version = "0.10.0-alpha.0"
+version = "0.10.0"
 
 [badges.travis-ci]
 repository = "gyscos/Cursive"


gyscos:
Hi! Thanks for the work!
Sorry I haven't had a look earlier, I'm just back from vacations and will check this now,

src/backend/puppet/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+                            return;
+                        }
+                    }
+                }
+            }
+        });
+    }
+
+    fn print_at(&self, pos: Vec2, text: &str) {
+        for (idx, c) in text.graphemes(true).enumerate() {
```
gyscos: Note that graphemes don't all have the same width (most asian characters are 2-wide).

src/backend/puppet/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+            let mut f = self.current_frame.borrow_mut();
+            let mut cpos = pos;
+            cpos.x += idx;
```
gyscos: Nitpicking: this could also be done without mutability with `let cpos = pos + (idx, 0);`

src/backend/puppet/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    ///
+    pub fn last_frame(&self) -> Option<&ObservedFrame> {
+        self.frames.front()
+    }
+}
+
+impl backend::Backend for Backend {
+    fn finish(&mut self) {
```
gyscos: `finish()` is called at the end, when the backend is about to be dropped and we want to "clean up" the terminal.
It's not called after each frame, you may instead be interested in `refresh()`.

njskalski:
Thanks! No need to apologize, I hope you had a good vacation! I will try to fix those, proceed with providing simple interface (as determined by needs of sly-editor) and we'll do the next round, ok?

PR #310, opened 2019-01-18
Test backend

njskalski:
I rewrote the test backend, I am starting slowly to write first tests ( https://github.com/njskalski/sly-editor/blob/master/src/file_dialog.rs ) figuring out what querying interface should look like. Any advice / constructive criticism appreciated.


src/backend/puppet/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+    },
+    effects: enum_set!(),
+};
+
+pub struct Backend {
+    inner_sender: Sender<Option<Event>>,
+    inner_receiver: Receiver<Option<Event>>,
+    prev_frame: RefCell<Option<ObservedScreen>>,
+    current_frame: RefCell<ObservedScreen>,
+    size: RefCell<Vec2>,
```
gyscos: Note that `Vec2` is `Copy`, so you could even just use `Cell`.

gyscos:
I think it may help to add an example, to show how the backend would be used.
It might even be useful to write this example _before_ writing the code for the backend itself, to help drive the API.

njskalski:
I am writing first tests in sly-editor, but haven't dedicated much time to it yet.

njskalski:
I updated interface, added some basic getters. I think it will be slowly incrementally evolving. let me know what you think.

What about adding serde to allow dumps of screens to files/load tests screens from files?

njskalski:
here are some first tests:
https://github.com/njskalski/sly-editor/blob/master/src/file_dialog.rs#L719

njskalski:
@gyscos I would appreciate some feedback even at this point. I plan to leave for entire March (no programming) so maybe we could merge even a limited version of this into dev branch? I think interface still needs "regex-with-colors" search (don't have the draft yet) and things like "get active window", but there's already a lot of code and I would prefer to get it reviewed before I start writing doc/coments to it.

njskalski:
I going to add comments and squash commits, so it's not such a mess to review.

gyscos:
Thanks again for all this!
I may have some time this weekend for an initial review.

gyscos:
Note: I (finally) landed - yet another - input refactor, which should simplify a bit the `Backend` trait. In particular, there is no need for a dedicated input thread anymore, now we are just polling the backend with `Backend::poll_event(&mut self) -> Option<Event>`.

njskalski:
Thanks. I am leaving for vacation, I will not be able to respond until April.

njskalski:
Hi. I am back from vacations. How are you?

gyscos:
Thanks for the updates!

I had a brief look, and I think it would help a lot to have some minimal example usage of this backend, to showcase how to write a small test for some dummy UI - maybe even one of the existing examples like `Select`.

> dumping/loading observed screens with serde, to make "screenshots" for comparison/diffing

This does sound cool, especially to automatically generate reports if a test fails.

> it would be great to be able to display "observed screen view" and then be able to draw something over it "transparently" (without filling background). Use case is simple: I load "screenshot" and use keyboard cursors to mark interesting part, and receive a "selector" to it that I can copy-paste to test.

A kind of pixel-art paint editor! Sounds very cool, but I can see it taking quite a bite of work.

> Next step would be "record/replay/diff interaction". Then I could write super simple tests like "I expect it to behave like usually when I do following steps", which would be great for finding side effects/breaking changes on the fly.

[I did something like that once at the shell level](https://github.com/gyscos/record/) (though without the diff mode), though a Cursive-specific one may be nice to have as well. Recording and replaying events+timestamps sounds simple enough, the diff view might take a bit more work. It'd be great in that it would let non-developers "write tests" by recording valid interactions and running the non-regression checks in CI.

njskalski:
> I had a brief look, and I think it would help a lot to have some minimal example usage of this backend, to showcase how to write a small test for some dummy UI - maybe even one of the existing examples like `Select`.

I uploaded one. At some point I might be able to clear up "BasicSetup" thing into something more streamlined, but I would prefer not to focus on it. All requirements will show themselves as I work on my project. This test is basically copy-paste from it.

> A kind of pixel-art paint editor! Sounds very cool, but I can see it taking quite a bite of work.

I might write it anyway. Do you think it would be a big change/is Cursive already ready to do transparent view? If it's not super expensive, I will write it anyway. I need tests, my project falls apart before assembly without them.

> [I did something like that once at the shell level](https://github.com/gyscos/record/) (though without the diff mode), though a Cursive-specific one may be nice to have as well. Recording and replaying events+timestamps sounds simple enough, the diff view might take a bit more work. It'd be great in that it would let non-developers "write tests" by recording valid interactions and running the non-regression checks in CI.

Smart diff (find out small shifts) would be hard. Simple diff (rule of the thumb) sounds easy.

njskalski:
I think I had somewhere a view that renders a captured screen, something that allows picture-in-picture, that could work as basis for pixel editor, but if you don't mind, I'd break the process into several smaller CLs.

gyscos:
> I might write it anyway. Do you think it would be a big change/is Cursive already ready to do transparent view? If it's not super expensive, I will write it anyway. I need tests, my project falls apart before assembly without them.

It already does! By default `StackView` adds a background to its layers, but with [`StackView::add_transparent_layer()`](https://docs.rs/cursive/0.12.0/cursive/views/struct.StackView.html#method.add_transparent_layer) you can add transparent views.

> Smart diff (find out small shifts) would be hard. Simple diff (rule of the thumb) sounds easy.

Yes, I think a git-like smart diff would be out of scope for now. A very simple mode that can highlight the diff area or something would already be pretty good to automate test reports.

njskalski:
ok, but how about this PR? Give me some actionable feedback so I can iterate towards landing.

src/backend/puppet/mod.rs
```diff
@@ -0,0 +1,173 @@
+//! Puppet backend
+
+#![warn(missing_docs)]
+#![allow(warnings)]
+#![warn(unused)]
+#![allow(bad_style)]
```
gyscos: We'll probably want to remove these before landing

examples/select_test.rs
```diff
@@ -0,0 +1,179 @@
...
+            BasicSetup {
+                siv,
+                screen_sink : sink,
+                input,
+                last_screen : RefCell::new(None)
+            }
+        }
+
+        pub fn last_screen(&self) -> Option<ObservedScreen> {
+            while let Ok(screen) = self.screen_sink.recv_timeout(Duration::new(0, 0)) {
```
gyscos: `recv_timeout` with an empty timeout is equivalent to `try_recv`.

examples/select_test.rs
```diff
@@ -0,0 +1,179 @@
...
+    use cursive::backend::puppet::observed::ObservedScreen;
+    use cursive::event::Event;
+    use std::cell::RefCell;
+    use std::time::Duration;
+    use cursive::event::Key;
+    use self::cursive::backend::puppet::observed::ObservedCell;
+
+    pub struct BasicSetup {
+        siv: Cursive,
+        screen_sink: crossbeam_channel::Receiver<ObservedScreen>,
```
gyscos: Is it really a sink? It seems it's a source (or a "stream") instead.

examples/select_test.rs
```diff
@@ -0,0 +1,179 @@
...
+
+        pub fn last_screen(&self) -> Option<ObservedScreen> {
+            while let Ok(screen) = self.screen_sink.recv_timeout(Duration::new(0, 0)) {
+                self.last_screen.replace(Some(screen));
+            }
+
+            self.last_screen.borrow().clone()
+        }
+
+        pub fn draw_screen(&self, screen: &ObservedScreen) {
```
gyscos: Does it need `self`? It might as well be a free function, or even something on `ObservedScreen` itself?

examples/select_test.rs
```diff
@@ -0,0 +1,179 @@
...
+                    if cell_op.is_some() {
+                        let cell = cell_op.as_ref().unwrap();
+
+                        if cell.letter.is_continuation() {
+                            print!("c");
+                            continue;
+                        } else {
+                            let letter = cell.letter.unwrap();
+                            if letter == " " {
+                                print!(" ");
```
gyscos: Do need to handle the space case separately?

src/backend/puppet/observed.rs
```diff
@@ -0,0 +1,585 @@
...
+
+impl ObservedScreen {
+    /// Creates empty ObservedScreen
+    pub fn new(size: Vec2) -> Self {
+        let contents: Vec<Option<ObservedCell>> = vec![None; size.x * size.y];
+
+        ObservedScreen { size, contents }
+    }
+
+    fn flatten_index(&self, index: &Vec2) -> usize {
```
gyscos: `Vec2` is copy, in general it's easier to pass-by-value.

src/backend/puppet/observed.rs
```diff
@@ -0,0 +1,585 @@
...
+    }
+
+    /// Returns a rectangular subset of observed screen.
+    pub fn piece(&self, min: Vec2, max: Vec2) -> ObservedPiece {
+        ObservedPiece::new(self, min, max)
+    }
+
+
+    /// Returns occurences of given string pattern
+    pub fn find_occurences<'a>(
```
gyscos: Could this be implemented for `ObservedPieceInterface`? It seems to use `self.min()` and `self.max()`.

gyscos:
Ok, I'm starting to see how it works. We'll probably just want to polish it a bit (add the missing docs and fix small details) and we'll merge it as a starting point.

At some point I think I'll update the example to isolate the "build the UI" part and the test part, to better show how it could be use to test existing UI code.

njskalski:
Updated to newest Cursive, fixed build errors and warnings. Now it's clean.

examples/select_test.rs
njskalski: yes, I keep printed space and None when there's nothing printed.

src/backend/puppet/observed.rs
njskalski: fixed it everywhere, thanks!

src/backend/puppet/observed.rs
njskalski: you are probably right, but for some reason it does not allow me to use index operator on `ObservedPieceInterface`.

njskalski:
OK I think it's ready for a (maybe last?) review, I am happy to follow up with consecutive comments in separate PRs.

gyscos:
Thanks again for all the work!

PR #314, opened 2019-01-28
Example program derived from linear.rs using decoupled UI

gbonnema:
Having time to spend I was toying with an idea the David Simmons offered in one of his blogs (https://cafbit.com/post/cursive_writing_terminal_applications_in_rust/). I took linear.rs and translated it into an mvc lookalike. Because I wanted to see keystrokes as well, I added 'q' and 'Q' to quit the example.

Let me know if you think it is worth showing, or not. Any way, all praise goes to David Simmons and all critics belong to me alone :-).

I renamed it mvc-linear to show the difference with linear.


gbonnema:
Hi, and thanks for the comments. I agree with most of what you say. I just started rust, so I don't get everything you say.

What I like about the blog is the way it is possible to separate model from view. If you want to see a example with more meat on the table, have a look at David's version of tupm: https://github.com/simmons/tupm.  The ui/controller/model separation is a lot better and the program is non-trivial and it probably reflects the setup a lot better than my example program. Maybe we can figure out an example that would have a better representation of the intention? 

P.S. The tupm program was written 2 years ago and is compatible to the original upm program (http://upm.sourceforge.net/). As David states though, don't use it for real: it's crypto part is really weak.

Also, it no longer compiles due to older crates it uses. I am working on that, but my knowledge of reqwest, mime and multipart is insufficient to solve it. The problems are only in one part (sync.rs) that is meant make a backup of the database in case the program corrupts it.

gyscos:
Hi again, and thanks for keeping this up to date!

I think this example might need to better showcase the improvements. In particular, we may want to find a scenario where having this kind of controller wrapping the event loop brings a clear benefit. For example, is there a situation where the controller can process messages in a unique way that `Cursive::cb_sink` couldn't support? That would probably be a good start.

If you want, we may jump on gitter.im to talk a bit about the design.

gbonnema:
Thanks for the reminder. 

My view on this example has changed: your remarks, which I at the time did not fully understand, are right. This example as is, is not useful.

However, I do think that an model-view-controller example would be a good thing. Just not this one. 
I would propose to close this one and abandon it. 

We could be opening a new issue for a generally acceptable model-view-controller setup that would have less  issues with the example.

Although I got distracted from working on this example, I did continue working on model-view-controller in Cursive. At the moment I am stuck. Also, I am pretty slow at adjusting stuff, especially if it's new to me, like rust.

If you want to we could agree on a time to discuss stuff in gitter. When we do, probably best if I send you what I have got so far so you can understand what I am trying to achieve.

gyscos:
> However, I do think that an model-view-controller example would be a good thing. Just not this one.
> I would propose to close this one and abandon it.

Sure! I think we should start with a better idea of the strong aspects of this design, so people reading the example will think "Wow! That's a cool design, it really helps a lot".

Maybe interacting with another "event loop", like a network service, could be such a scenario. For example, we have some TCP server that adds some offset to each `u8` received and echos this back. The UI would control the offset, and shows logs what's going on - and maybe it could also run a test connection. This way, we have:
* Model: the offset, and logs of past events
* View: A slider/editview for the offset, and some scrolling list for the logs

That's just a simple example I just came up with, but we can also brainstorm and look for something better.

I think the `Cursive` object is already working as a controller most of the time, handling input and callbacks. Splitting model and view, on the other hand, seems like a good low-hanging fruit.

> If you want to we could agree on a time to discuss stuff in gitter. When we do, probably best if I send you what I have got so far so you can understand what I am trying to achieve.

Sounds good! My time schedule is pretty erratic these days, so just drop a message whenever you have time, I may be able to answer :)

gbonnema:
I found your tcp-program and I am studying it at the moment. First off: your knowledge and skill are way beyond my level. That probably means I will have to take my time to understand what is happening here.

Do you want to open a new issue or PR and close this one? It seems the tcp example is already part of master? Let me know.

gyscos:
Ha ha thanks, but don't underestimate yourself! I'm merely more familiar with cursive itself, so it's a bit easier for me.

I'll document a bit more the tcp example to make it easier to read and understand.

We can use this thread to converge on a better understanding of what kind of example we need, and see if the tcp example fits this or if we need something more.

gbonnema:
I *do* hope this rant does not put you off. Don't be afraid to let me know, if this *not* what you have in mind.

I had a chance to look things over. Also, I had a chance to look over what I set out to do. I suspect it may be a good time to make my ideas about what I have in mind more explicit. That way, we can see if we are on the same page or we have totally different idea's.  Or even, if the idea's are feasible. 

One consequence is that I now understand why we should be using `mpsc` or `crossbeam` and `bus` in stead of `cb_sink`.  Using `cb_sink` requires being *inside* the UI.

The question is also whether we can show all this in a simple example. Tbd.

**MVC model**
I have the following in mind :

1. The model and UI are independent. 
2. The coupling between model and UI is as loose as possible. 
3. The application would connect model and UI through independently creating them and having them execute.

Ad 1. *The model and UI are independent* 
This means there is no "knowledge" of the UI in the model *and* no knowledge of the model in the UI. The interface between model and UI should bridge that gap promoting loose coupling and hopefully maximizing coherence.

As a consequence, if I were to use your tcp model (say as a library), I could use it for a gui with the same ease as for a tui and for both I could choose my own engine (GTK vs QT or Cursive vs .....).  This precludes the use of `cb_sink`.

Ad 2. *The coupling between model and UI is as loose as possible* 
The interface between model and UI should reflect only the aspects that UI and model have in common. So the model should relay *how* to show the data:
a. graph, text or binary
b. text: single line, multiple line
c. text: secret or public
d. text: key field or not
et cetera. 

The model should also relay *what* to show. This would be the contents / data.

Ad 3. *The application would assemble to parts*

The application would be the controller. It would start the ui and connect to the model. It would not do *any* functions of the model or the ui. Connecting UI and model should enable communication between the two.

**Design**
The application programmer will automatically separate model and view in a way that is most independent. One could completely replace the UI without touching the model and vice versa.  Another advantage would be to have an implementation in TUI *and* in GUI depending on environment or choice of the user, without adapting the model. The interface of *what* to to show and *how* to show would remain the same.

**Use case**
Imagine a use case where we would run one UI thread for 2 different UI's. Like when testing a new version of the UI and comparing the results with the old UI. One would have to run Cursive in each thread and have the model notify both in case of change of contents. Or one gui and one tui both showing different aspects of the data.  The model would notify the ones subscribed no matter which ui changes the data. 

If you imagine the tcp example where the model assumes the role of a server, you could imagine what I mean.

**Implementation**
***metadata***
Describing the fields and their characteristics isn't too hard. Just describe each field in a struct and have all the fields in an array and send that over to the UI on request.

***sending data***
Describing the contents I find a bit harder. I *did* decide to have all data in unicode (as we are showing on screen and in text), which simplifies things. The difficult bit is how to organize the data with respect to the meta data.

***updating data***
Describing a change shouldn't be too hard if you include before and after image of the data so the model can judge the change and either approve and apply or refuse. As the model reads all the udpate  requests from the message queue, there should be no data races. It does provide the ability to refuse based on stale data (before image is no longer actual).

***organizing message queue***
This is the hardest part (to me). We need a request and response channel per subscriber using `mpsc` or `crossbeam`. Also, we need the have the subscriber in a notify channel using a `bus` with broadcast abilities. 

***miscelaneous***
Finally, I am not sure whether we need locking. If all goes through messaging queues, Isn't the locking taken care of by OS and the compiler?

gyscos:
Thanks for the explanations! Indeed, I was missing the main point: to make the model independent from the UI.

So rather than directly communicating with the UI, the model should surface "events" that the controller then translates to UI actions. In the other direction, though, do we need the UI to be independent from the model?

Let's say the model defines multiple actions/events, represented as methods:

```rust
impl Model {
    pub fn set_offset(&self, new_offset: u8);
    pub fn get_logs(&self) -> &[LogEntry];
}
```

There are a few ways to connect this to the UI:
* Solution A: Directly give the UI a pointer/`Arc` to the model, and let it call the method when it needs.
* Solution B: Have the UI emit "UI events" through some channel, and in the controller translate these events into method calls on the model. To send back values from the model to the UI, we have again several options:
    * Solution B.1: The "event" message from the UI includes a way to get something back, like `eventual::Future`.
    * Solution B.2: There is another channel dedicated to sending data back to the UI.

The last option B.2 is the "purest" one, in that it uses channels all the way, but it's also the most annoying to set up as you mentioned: you need to give a request and a response channel to every part of the UI that needs it. You could wrap both in a struct you could clone and pass around, but at this point you're basically creating a new intermediate model in a situation like solution A.

So overall, a solution might be:
* Define model events: `enum ModelEvent { ... }`
* Define the model as `struct Model`, taking a `Sender<ModelEvent>` during construction.
    * The model defines possible actions as methods, either as intrinsic `impl` or as a trait implementation.
* Define the UI, taking `Arc<Model>` during construction.
    * When some input is received, the UI triggers the corresponding action by calling the method.
    * Alternatively, the model takes a `Arc<MiddleManModel>` instead, which forwards call to/from the actual model in _some_ way (either by embedding a `Arc<Model>` itself, or by using channels).
        * This has the result of keeping the UI independent from the `Model` - it now only depends on the controller's `MiddleManModel`. If `Model` is subject to change, this may help compatibility.
* Define the controller, which listens to the `ModelEvent` channel and translate these events into UI callbacks that it sends back to `cb_sink`.
* Start the model in its own thread.
* Start the controller in its own thread.
* Start the UI in the main thread (many UI libraries prefer to be started from the main thread).

gbonnema:
I hope you are not getting impatient. It's taking me some time to work something out. I do have a basic setup, but with several issues I haven't tackled yet.

gyscos:
> I hope you are not getting impatient.

Ha ha far from it! I'm currently moving to a new place so I understand being busy with other things ;)

Don't hesitate to ask if you have any question or doubt!

gbonnema:
I committed to a half finished mvc example program. Also, created a thread.rs to show what happens if you use Cursive in a non-main thread (do not do this at home or in production :-)).

The program does show a window -- sort of -- but it is smallish and does not look nice. Also, I don't retrieve data yet.

I am wondering whether this is the path to proceed on. What I did:

1. I created a controller (which is the main function). It starts the ui and the model and does nothing else.
2. The ui uses a premade channel that it receives in the model struct. The model uses a response channel prepared by the ui.

As you see, I deviated from our design, in that the controller does nothing more than introduce the actors. Communication is independent. I used the analogy of a friend introducing a stranger to you. The stranger might send you a message and leave a return address to which you can respond. This is what I did in the program. The controller introduces the ui to the model. The ui then proceeds to send a message and includes a return address.

I am pretty sure the setup is not efficient, but for the time being a proof of concept is the first step.

My next step will be to show data as well. And finally, I should *really* make the window look nice(r than it does now).

Kind regards, Guus.

gbonnema:
oh, and to be complete: it's more code than I expected, so probably some of it can be refactored out into a crate. But as I said: proof of concept for now.

gyscos:
Running the UI in a new thread can often cause a bunch of problems - some libraries usually assume they run on the main thread and are confused when they aren't. It _might_ work on ncurses, but I haven't tested it much.

In addition, you need to keep the main thread "waiting": otherwise, when `main()` returns, it kills all outstanding threads, so the app will effectively terminate instantly - sometimes before even initializing ncurses, sometimes in the middle of it.

gbonnema:
hmmm, That *is* an idea. I had main return immediately, didn't even consider it. But anyway, the `thread.rs` was just a tryout. I my comment I was talking about `mvc.rs`. But I guess you guys have **4th of july** to think of, right?

gbonnema:
Quick question. When running rustfmt from the root directory of Cursive, I got warnings on not using the nightly.
image
Is it required or even advisable to use nightly? If so, can I do so on Cursive project only?

gbonnema:
Question: I do a callback from a button, but the compiler complains that the &mut self does not have a static lifetime. How does one solve that? I will commit the change with the offending statements in comments. The code looks like this:

image

Do you think you could point me in the right direction?

Kind regards, Guus.

gbonnema:
This morning I had an idea: I remember you recently introduced user_data. I could add a pointer to the model in there and use the pointer as a selfish pointer. I do realise I will have to use `rc` and maybe even `refcell`. I will have to look it: both are new to me. Either that, or I store model in the user_data only and not in the `Ui` struct.

If you feel there is a better solution, let me know.

gbonnema:
Some of the checks failed (travis). I suspect it has to do with the use of user_data. It points to the stable version of rust not being able to handle the feature Maybeuninit:
```use core::mem::MaybeUninit; // Rust 1.36+ required```

The nightly rust version seems to be Ok with it.

**Comment by gyscos** [comment]
> Quick question. When running rustfmt from the root directory of Cursive, I got warnings on not using the nightly.

Ah, that's possible, I only use the nightly formatter.

> Question: I do a callback from a button, but the compiler complains that the &mut self does not have a static lifetime. How does one solve that?

When you define the new `Dialog`, you cannot keep any reference to `self`, since the `Dialog` may outlive it.
The usual way to access external data is with `Rc` (or `Arc` if multithreading). This way, every closure has its own `Rc` and can access the `Ui` independently.

The other way is through `Cursive` itself. It can work if what you're trying to access is another view in the tree - `Cursive::call_on_id` or `find_id` are here for that. Or, it can also work if you stash the thing to access in the `user_data`, as you noticed.

> Some of the checks failed (travis). I suspect it has to do with the use of user_data. It points to the stable version of rust not being able to handle the feature Maybeuninit:
use core::mem::MaybeUninit; // Rust 1.36+ required

Seems to be a problem with  a transitive dependency: the `array-macro` crate bumped the minimum rustc version in a patch release: https://gitlab.com/KonradBorowski/array-macro/issues/2
It should be fixed now (they fixed the crate we depend on to use an older version of `array-macro`).

**Comment by gbonnema** [comment]
the source code of `views/linear_layout.rs` has active `log::debug!` statements in it. Those all appear in my toggle_debug screen and produce output very fast: did you commit that by accident or intentional?

**Comment by gyscos** [comment]
Ah - I guess I forgot about these. They can probably be commented out by now.

The ideal solution would be to keep them, but setup the debug window to use different log levels for different modules (local crate vs cursive vs other dependencies...).

**Comment by gbonnema** [comment]
> The ideal solution would be to keep them, but setup the debug window to use different log levels for different modules (local crate vs cursive vs other dependencies...).

Ahhh, that would be nice. I will check the `log` crate, to see if it is already an option.

**Comment by gbonnema** [comment]
> The usual way to access external data is with `Rc` (or `Arc` if multithreading). This way, every closure has its own `Rc` and can access the `Ui` independently.
> 
> The other way is through `Cursive` itself. It can work if what you're trying to access is another view in the tree - `Cursive::call_on_id` or `find_id` are here for that. Or, it can also work if you stash the thing to access in the `user_data`, as you noticed.

Thanks for the options. I consider combining the `user_data` solution with sending a signal. That way a button callback never has to "do" anything, except send a signal that it was pressed. The Ui object will handle business. I will let you know how that works out.

**Comment by gbonnema** [comment]
Signalling the Ui object from the button seems to work. However, updating the data after a fetch, does not. I realize I need to fill the content of the EditView after each refresh. So that will be the next step: finding out how to do that. I have already given the EditViews an `id`. Now I just need to find the event where I can do a call_on_id.

Will keep you posted.

**Comment by gbonnema** [comment]
I committed a working version.

That said, it looks like *not* an example program. Also, in 1 file, it is not easy to see the distinction between the general stuff of the model and the specific stuff of the model (custlist as an example). Creating a simple, cleanly designed mvc program (like mr Simmons did in `tupm`) looks a lot easier than it is xD.

And finally, the code has a *lot* of duplication. At school it would get bullied for being bloated. 

I also remarked, that I cannot see the difference between a button `enabled` and `disabled`. Could be my terminal emulator of course.

**Comment by gbonnema** [comment]
I wanted to use the `execute_get_next` etc functions that return a callback, which seem to work for the buttons in the dialog, for a global callback. They appeared to do nothing. Even when I copied the routine to return an `FnMut` function it didn't work.

When I replaced the call to the `execute_*` function by direct code, it *did* work. Odd no?

P.S. the global shortcuts are PgUp and PgDn (I was getting fed up with the buttons).

**Comment by gyscos** [comment]
> EDIT: unfortunately, log only filters per log level. It looks like env_logger can, but we already have a logger installed in Cursive, so that won't be an option. I decided to set the max log level (info).

Yeah we'd need to add this feature to Cursive's own logger.

> When I replaced the call to the execute_* function by direct code, it did work. Odd no?

`execute_get_next` doesn't seem to do anything by itself: it just returns a boxed callback that does something.
So if you just call `Ui::execute_get_next();` when a key is pressed, it will create this callback, but not run it, so nothing happens. To run the callback you create, you could do:
```rust
        self.siv.add_global_callback(Key::PageDown, |s| {
            Ui::execute_get_next()(s);
        });
```

Or you could as well replace the entire closure with the result from `execute_get_next`:

```rust
        self.siv.add_global_callback(Key::PageDown, Ui::execute_get_next());
```

gyscos:
> That said, it looks like not an example program. Also, in 1 file, it is not easy to see the distinction between the general stuff of the model and the specific stuff of the model (custlist as an example). Creating a simple, cleanly designed mvc program (like mr Simmons did in tupm) looks a lot easier than it is xD.

Designing minimal examples can certainly be an art sometimes :)

If the example warrants it, it could be split into multiple files like the `mines` example. Or it could also define separate modules in the same file if it's small enough.

gbonnema:
Thanks for the advice, I am not happy with the current setup: too much specific code for it to be useful.
Let me work on it a bit more. 

(P.S. not sure how much of a time diff we have, but here it is about 23:15

gbonnema:
Could I ask a favour? Would you want to have a look at the code I have created in https://gitlab.com/gbonnema/mvc-rs (not github, and you need to logon to find it). The code that I have a problem with is `with_id` in combination with `call_on_id` where I suspect the call_on_id cannot find the EditView I need.

The code lines are in respectively `examples/simple.rs:248` and `examples/simple.rs:405`.

```237         let record = match &self.db_record.record {
238             None => {
239                 self.buttons_refresh();
240                 return;
241             }
242             Some(rec) => rec,
243         };
244         let data = record.data();
245         log::warn!("fields_refresh: data = {:?}", data);
246         for (id, value) in data {
247             log::warn!("id = {}, value = {}", id, value);
248             self.siv.call_on_id(id.as_str(), |view: &mut EditView| {
249                 view.set_content(value);
250             });
251         }
252         self.buttons_refresh();

```

```
396                     let text = format!("{}: ", field.label().clone());
397                     let id = format!("id_{}", field.name().clone());
398                     let mut hview = LinearLayout::horizontal();
399                     let fld_data: String = match rec.data().get(field.name().as_str()) {
400                         None => "".to_string(),
401                         Some(s) => s.clone(),
402                     };
403                     hview = hview
404                         .child(TextView::new(text).fixed_width(maxlen))
405                         .child(EditView::new().content(fld_data).with_id(id).min_width(10));
406 
407                     view = view.child(hview);
```

I added several `log::warn!` statements before being sure this was the problem (it could have been anything). I just have no idea why it would think the type is not an EditView?

gbonnema:
Sorry, closed by accident.

gbonnema:
> Could I ask a favour? Would you want to have a look at the code I have created in https://gitlab.com/gbonnema/mvc-rs (not github, and you need to logon to find it). The code that I have a problem with is with_id in combination with call_on_id where I suspect the call_on_id cannot find the EditView I need.

Yeah ...... what can I say .... ? Just a bug. I had the glorious idea to prefix the id, but forgot to do it.
My apologies. Hope you didn't spend any time on it, embarassed .....

gbonnema:
@gyscos I have created this crate `mvs-rs` in gitlab (https://gitlab.com/gbonnema/mvc-rs), where I extract as much generic code as possible into the crate and have the example program (`examples/simple.rs`) be as simple as I think it can be. But it still is much more than 100 lines.

I would really love to have a way to separate user interface and the model cleanly. My newbishness in this kind of design however, gets in the way.

Could you have a look and see if anything jumps out? Maybe you can improve on the design or maybe I am just taking the wrong "prong" somewhere. Let me know if you can, and then .... just take your time. I wouldn't want to impose or rush you. If you cannot, let me know too.

gyscos:
Thanks for the continued support - and sorry I haven't had time to give it a better look yet!

(As for the gitlab account - any reason the repo isn't public?)

Let's recap the minimum we need for an example:
* A process, independent of the UI - exposes signals and actions.
* A UI, independent of the process - exposes signals and actions.
* A controller to connect the signal from one side to actions on the other side.

In an asynchronous scenario where the process run in its own thread (like the tcp example I started), signals may be channels (and actions methods).
A synchronous scenario is also possible, where the process is mostly reactive (for example a tic-tac-toe game, or the minesweeper example), in this case signals can just be callbacks.

The controller would differ a bit between these two situations, so we might want to have an example for each. In both cases, the UI and the process each will be defined in their own module, so it's clear they are independent.

gbonnema:
> Thanks for the continued support - and sorry I haven't had time to give it a better look yet!

Take your time. I do too. I will wait for your judgement then. 
> 
> (As for the gitlab account - any reason the repo isn't public?)

Gitlab has 3 levels: private (only on invitation), semi-public (when logged on to github), public (no logon necessary).

I didn't want my repo to be found by any arbitrary user (it's not ready for that) so I made it semi-public.

> 
> Let's recap the minimum we need for an example:
> 
>     * A process, independent of the UI - exposes signals and actions.
> 
>     * A UI, independent of the process - exposes signals and actions.
> 
>     * A controller to connect the signal from one side to actions on the other side.
> 

I had the generic part of the model and of the ui in the src directory of mvc. Most of it in mvc.rs and some cursive specific stuff in cursive.rs. It uses mprc to create channels, with the ui creating a return channel for the responses. The example source uses these structs and implements the traits to do this communcation. 

The controller I did not program separately, but had the `main` function be the controller. Multiple threads seems unavoidable in this setup.

> 
> In an asynchronous scenario where the process run in its own thread (like the tcp example I started), signals may be channels (and actions methods).

I think I had everything be a message over a channel. So only the model executes methods from the model and only the UI executes methods from the UI. I am not sure I understand completely what you mean by process: do you mean the model?

> A synchronous scenario is also possible, where the process is mostly reactive (for example a tic-tac-toe game, or the minesweeper example), in this case signals can just be callbacks.

This one, I do not see (in my head). Maybe I am too much hooked on using channels and thus on threads waiting for a message.

> 
> The controller would differ a bit between these two situations, so we might want to have an example for each. In both cases, the UI and the process each will be defined in their own module, so it's clear they are independent.

Okay, I may have gotten this wrong. We will know when you have had a chance to look at what I cooked up. Again: no pressure. Take your time, I know you must have a lot of other stuff to look at. Also, my coding may be a bit wordy in comparison to what professional programmers produce (especially the ppl that know Rust well).

Kind regards, guus.

gyscos:
> I think I had everything be a message over a channel.
What I meant is that the controller has knowledge of both the model and the UI, so it _could_ call public methods on either - for instance if it had a `Arc<Model>` it wouldn't need to use channels to trigger actions on the model, only to receive events. When to use channels vs direct methods depends on the perf requirements and the timing (can "events" be processed anytime, or in a strict schedule?).

> I am not sure I understand completely what you mean by process: do you mean the model?
Yes - sorry I mixed up the two terms.

> This one, I do not see (in my head). Maybe I am too much hooked on using channels and thus on threads waiting for a message.

Imagine a 2-players tic-tac-toe: the model is entirely synchronous and never triggers event by itself. Instead, it directly "reacts" to UI-driven events. In this case the controller doesn't have to listen to messages from a model thread; it can directly call methods on the model. And it doesn't even _need_ to listen to a channel from the UI thread; instead, the UI can store callbacks (`Box<Fn()>`) rather than sending out messages. The controller would set these callbacks to call the model.
This results in a low-overhead single-threaded synchronous implementation where UI and model and entirely independent, and the controller just connects the two.

gbonnema:
> This results in a low-overhead single-threaded synchronous implementation where UI and model and entirely independent, and the controller just connects the two.

Yes, that seems to be a nice solution for single threaded programs. And thank you very much for your explanation.

> > I think I had everything be a message over a channel.
> 
> What I meant is that the controller has knowledge of both the model and the UI, so it _could_ call public methods on either - for instance if it had a `Arc<Model>` it wouldn't need to use channels to trigger actions on the model, only to receive events. When to use channels vs direct methods depends on the perf requirements and the timing (can "events" be processed anytime, or in a strict schedule?).

This one, I have been mulling over in my head. Assume method calls are better than messages in some circumstances. Who _should_ do method calls? With the criterium of maximum coherence and minimal coupling, the controller seems to function as an intermediary, knowing nothing of the function of the UI or the model.  Like in your single threaded proposal. So does our controller, it introduces and stands back. Let the interface between model and UI do the work.

Still: who _should_ call the methods? I suspect the answer is the same as for messages over channels. The model should create an interface that the user interface can use.  That way maintenance of model related stuff is *all* in the model, while only the user interface related stuff remains in the UI. Additionally, the controller does not add any coupling to the interfaces.

Getting the mvc with messaging right, it should not be too hard to get the methods based mvc right.

So the first step is: getting mvc-messaging right. And *that* seems to be a *hard* question.

Let me know when you had time to look over the sources. I divided stuff into an mvc crate and the example program (simple). I know I have to move some more stuff from the example program to the mvc crate, it will just take time to get it right.

The question now is : do you think the current way of doing things is the way to go?

gyscos:
Closed by mistake.

gbonnema:
Hey gyscos, this branch is stale and I guess it will not improve as I cannot spend time on it. I am closing it now as I don't see this going anywhere. If I want to do something similar or different, I will open a new pull request. Thank you for your patience.

PR #321, opened 2019-03-03
Reduce the title drawing cut-off point by 2 characters.

chrisvest:
Panels and dialogs have borders with corners.
Previously, the title would draw over the border corner, including its own end-points.
Now the cut-off point is such that we do not draw the title if the right-hand end point and the border-corner do not also fit.

### DIFF
--- a/src/views/dialog.rs
+++ b/src/views/dialog.rs
@@ -461,14 +461,15 @@ impl Dialog {
     fn draw_title(&self, printer: &Printer<'_, '_>) {
         if !self.title.is_empty() {
             let len = self.title.width();
-            if len + 4 > printer.size.x {
+            let spacing = 3; //minimum distance to borders
+            let spacing_both_ends = 2 * spacing;
+            if len + spacing_both_ends > printer.size.x {
                 return;
             }
-            let spacing = 3; //minimum distance to borders
             let x = spacing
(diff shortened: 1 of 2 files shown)


gyscos:
Hi, and thanks for the work!

Indeed, the spacing moved from 2 columns to 3, and not all part of the code were updated for that :(

Regarding the failing CI, for some reason rustc 1.33 has been failing on windows, not sure why yet...

gyscos:
A solution could be to print the title in a sub-printer (offset and cropped).

gyscos:
As for a more generic method to add to `Printer`, I think supporting multiple use-cases would make the function a bit too complex. I imagine the signature would be something like:

```rust
fn print_title(&self, start: impl Into<Vec2>, len: usize, alignment: HAlign, left_filler: &str, left_border: &str, right: border: &str, right_filler: &str, title: &str);

printer.cropped((2,0)).offset((1,0)).with(|printer| {
    printer.print_title((0,0), printer.size.x, HAlign::Center, "-", "┤", "├", "-", title);
});
```

But I don't think it's currently justified.

gyscos:
Though taking a second look, we do have `print_box`, so would not be so far off, and wouldn't require this level of genericity.

PR #324, opened 2019-03-11
Make StackView::draw_bg faster.

chrisvest:
Speed up the printing of the background in the StackView by letting Printer::print_hline cache its buffer of space characters.

I found this while profiling the rendering of my program. It's a smallish gain. And perhaps a bit too hacky to merge, I don't know. I'm still quite new to Rust, so I welcome any feedback on this you might have.

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -10,6 +10,7 @@ use crate::vec::Vec2;
 use crate::with::With;
 use enumset::EnumSet;
 use std::cmp::min;
+use std::sync::{Arc, Mutex};
 use unicode_segmentation::UnicodeSegmentation;
 use unicode_width::UnicodeWidthStr;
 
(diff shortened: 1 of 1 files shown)


gyscos:
Thanks for the PR!

I'm still wondering what the best solution would be here. Ideally we'd want to avoid allocation in all cases, not just when printing spaces.

One direct extension of this PR would be to keep a `Rc<RefCell<HashMap<&str, String>>>` and store a buffer for each character used in `print_hline` (No need for `Arc<Mutex<_>>` since we're single-threaded here).

Another solution would be to use some buffering: if `Backend::print_at` is cheap, we could call it in a loop and print each character individually. We don't do it currently because it would mean more output to send to the terminal (a lot more control codes to move the cursor by 1 character), but a buffering solution could solve this.

Yet another solution would be to refactor the `Backend` trait to make it more efficient, maybe with a `repeat` parameter in `print_at`. This way, most backends could implement it as a single move and then multiple prints, while `BearLibTerminal` would just print each character individually (but bandwidth is not a problem there).

chrisvest:
I'll give it a try with a `print_at_rep` method and see how it is.

chrisvest:
I think the approach in #326 looks and works a lot better, so I'll close this.

PR #325, opened 2019-03-12
Use a single hashmap lookup instead of two for resolving ncurses colours.

chrisvest:
It's not so easy to quantify the improvement, but these calls show up a little bit less in profiles when rendering lots of text that colours a lot. Think syntax high-lighting and other forms of colour coded text heavy screens.


gyscos:
Awesome, thanks!

PR #326, opened 2019-03-12
Add Backend::print_at_rep and make Printer::print_hline use it.

chrisvest:
This avoids a string allocation in `print_hline` and makes it faster, presumably for all backends.
This speeds up the rendering of the background in StackView.

This supersedes #324, since it not only looks simpler when each backend is taken in isolation, but also appears to make the rendering even faster.

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -246,7 +246,7 @@ impl<'a, 'b> Printer<'a, 'b> {
         let start = start - self.content_offset;
 
         // Don't write too much if we're close to the end
-        let repetitions = min(width, (self.output_size.x - start.x)) / c.width();
+        let repetitions = min(width, self.output_size.x - start.x) / c.width();
 
         let start = start + self.offset;
         self.backend.print_at_rep(start, repetitions, c);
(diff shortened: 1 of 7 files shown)


src/printer.rs
```diff
@@ -248,11 +248,8 @@ impl<'a, 'b> Printer<'a, 'b> {
         // Don't write too much if we're close to the end
         let width = min(width, (self.output_size.x - start.x) / c.width());
 
-        // Could we avoid allocating?
-        let text: String = ::std::iter::repeat(c).take(width).collect();
-
         let start = start + self.offset;
-        self.backend.print_at(start, &text);
+        self.backend.print_at_rep(start, width, &c);
```
gyscos: I think you don't need to reference `c`, it's already a `&str`.

src/printer.rs
```diff
@@ -248,11 +248,8 @@ impl<'a, 'b> Printer<'a, 'b> {
         // Don't write too much if we're close to the end
         let width = min(width, (self.output_size.x - start.x) / c.width());
```
gyscos: I think the computation here is wrong: it should be `let repetitions = min(width, (self.output_size.x - start.x)) / c.width();`

chrisvest: Yeah, I think you're right.

chrisvest:
@gyscos Thanks for the review. I think I've addressed all of your comments.

gyscos:
Thanks again!

PR #327, opened 2019-03-12
Avoid iterating graphemes unless we have to.

chrisvest:
Iterating graphemes with `text.graphemes(true)` is somewhat expensive due to the complexity of unicode.

However, we were doing it _twice_ in the `Printer::print` method in order to truncate the input text to fit the content area.

This changes that method to track the width of the input text, and only do the grapheme iteration when truncation is actually necessary.

This seems to the very helpful for my case where I print many small strings. I get roughly a 40% speed-up in refresh rate from this.

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -129,46 +129,55 @@ impl<'a, 'b> Printer<'a, 'b> {
             return;
         }
 
-        let text_width = text.width();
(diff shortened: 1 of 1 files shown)


gyscos:
Woo, thanks a lot for the work!

PR #329, opened 2019-03-21
Add a sort method to SelectView, to easily sort all contained items lexicographically by their label.

chrisvest:
I was populating a `SelectView` with files from `read_dir`, which returns them in whatever order the OS feels like, and thought it'd be convenient if I could just tell the `SelectView` to sort itself.

I tried to use `Vec::sort_by_key` at first, but apparently the lifetimes around the closure got too complicated for me to deal with, so I implemented `Ord` (and friends it insisted on bringing along) for `Item` instead.

### DIFF
--- a/src/views/select_view.rs
+++ b/src/views/select_view.rs
@@ -389,7 +389,7 @@ impl<T: 'static> SelectView<T> {
     /// Note that this does not change the current focus index, which means that the current
     /// selection will likely be changed by the sorting.
     /// This sort is stable: items with identical label will not be reordered.
-    pub fn sort(&mut self) {
+    pub fn sort_by_label(&mut self) {
         self.items.sort();
     }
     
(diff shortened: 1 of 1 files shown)


gyscos:
Thanks for the work!

I'm wondering what the best solution is here. We probably want to support sorting on more than just the labels, but something based on the values as well. Maybe provide `fn items(&mut self) -> &mut [impl ItemTrait];` with some kind of trait? Or maybe just a `sort_by`/`sort_by_key` method with a closure that takes both the label and value?

chrisvest:
@gyscos I renamed the initial sort method to sort_by_label, and then added sort methods that mirror those of Vec, so the API looks familiar. It was also easier to do it this way: the lifetimes and ownerships are more straight forward, I feel.

What do you think of the changes now?

gyscos:
Awesome! Thanks a lot for the work!

PR #332, opened 2019-03-29
Ensure `Cursive::step()` does not return until FPS timeout elapses if idle

### LINKED ISSUE #331: set_autorefresh pattern leads to unnecessarily high redraw frequency than set_fps used to
Bug report / suggestion

# Problem description

With 0.10, it was possible to write your own mainloop with interleaved computation, like:

```rust
fn main() {
    let mut state = State { ... };

    let mut window = Cursive::termion();
    window.set_fps(2);

    loop {
        state.update(); // Update the application state

        // Update the UI according to the new application state
        window.call_on_id(view_1_id, |view_1| update_view_1(view_1, &mut state));
        window.call_on_id(view_2_id, |view_2| update_view_1(view_2, &mut state));

        // Draw the updated UI, and wait for frame to elapse or new input event
        window.step();

        if !window.is_running() {
            return;
        }
    }
}
```

which achieves the following things:

1. Mutating the application state happens on the single "draw" thread, instead of needing to pass it back and forth between a "draw" thread that blocks on `Cursive::run()` and a "state" thread that updates it.

1. By capping the FPS to 2, the application state is only updated every 0.5 seconds when idle.

1. `Cursive::step()` returns immediately when an input (eg keypress) is available which causes the loop to re-run immediately. If the loop had its own separate `std::thread::sleep(<500 ms>)` then every keypress would appear unresponsive for upto 500ms.

But with 0.11's `set_autorefresh`, the default "FPS" is set to about [1 / 30ms](https://github.com/gyscos/Cursive/blob/0.11.0/src/cursive.rs#L827), which leads to excessively frequent updating of the state. Since updating the state also updates the UI, this means the application redraws constantly and is never idle, which leads to higher CPU usage.

Would it be possible to bring `set_fps` back? `set_autorefresh(true)` can continue to default the value to 1 / 30ms, but `set_fps` could let it be customized like before.

# Environment

* Backend used: termion
* Cursive version: 0.11.1

Arnavion:
Fixes #331

### DIFF
--- a/src/cursive.rs
+++ b/src/cursive.rs
@@ -809,8 +809,11 @@ impl Cursive {
     /// Useful if you need tighter control on the event loop.
     /// Otherwise, [`run(&mut self)`] might be more convenient.
     ///
+    /// Returns `true` if an input event or callback was received
+    /// during this step, and `false` otherwise.
+    ///
     /// [`run(&mut self)`]: #method.run
-    pub fn step(&mut self) {
+    pub fn step(&mut self) -> bool {
         // Things are boring if nothing significant happened.
         let mut boring = true;
 
@@ -820,7 +823,7 @@ impl Cursive {
             self.on_event(event);
 
             if !self.running {
-                return;
+                return true;
             }
         }
 
@@ -830,7 +833,7 @@ impl Cursive {
             cb.call_box(self);
 
             if !self.running {
-                return;
+                return true;
             }
         }
 
@@ -845,7 +848,6 @@ impl Cursive {
                 .unwrap_or(false)
         {
             // We deserve to draw something!
-            self.boring_frame_count = 0;
             self.refresh();
         }
 
@@ -853,10 +855,14 @@ impl Cursive {
             std::thread::sleep(Duration::from_millis(INPUT_POLL_DELAY_MS));
             self.boring_frame_count += 1;
         }
+
+        !boring
     }
 
     /// Refresh the screen with the current view tree state.
-    fn refresh(&mut self) {
+    pub fn refresh(&mut self) {
+        self.boring_frame_count = 0;
+
         // Do we need to redraw everytime?
         // Probably, actually.
         // TODO: Do we need to re-layout everytime?


gyscos:
Wow - thanks for the PR!

I think `Cursive::step()` doesn't need to be synchronized with the screen refresh rate. On the other hand, having it return consistently~ish allows a more predictable update cycle, especially when varying the refresh rate to fit the current views (enable auto-refresh when there's a dynamic view, disable it otherwise, use high FPS when a progress bar is visible but low FPS when we're just updating logs, ...).

This is especially true when there is no autorefresh ("0 fps"), as we might still want an update cycle.

If we really want to synchronize with the refresh rate, one solution might be to have `step` return a bool or some flag to indicate that a refresh was triggered.

Arnavion:
Right, I figured you might want that. Yes, I'm happy to move the "FPS" mechanism to the caller of `step()`, and yes then it needs to return a flag indicating whether it had a "boring" event or not before it returned. The caller needs that to decide whether to call `step` again immediately or update the state and UI first.

I'll update the PR in a few hours.

Arnavion:
Updated to have `Cursive::step()` return `!boring`

Also, for this comment from the issue:

>You'd need to trigger a new refresh, possibly by sending a dummy callback through `Cursive::cb_sink`, but then the next call to `step()` will also refresh the screen, which will cause another state update...

I made `Cursive::refresh()` public so that a custom mainloop can look like:

```rust
    loop {
        state.update(); // Update the application state

        // Update the UI according to the new application state
        window.call_on_id(view_1_id, |view_1| update_view_1(view_1, &mut state));
        window.call_on_id(view_2_id, |view_2| update_view_2(view_2, &mut state));

        // Draw the updated UI
        window.refresh();

        // Wait for frame to elapse or new input event
        let timeout = crossbeam_channel::after(std::time::Duration::from_millis(500));
        while let Err(crossbeam_channel::TryRecvError::Empty) = timeout.try_recv() {
                if window.step() {
                        break;
                }
        }

        if !window.is_running() {
            return;
        }
    }
```

gyscos:
Awesome, thanks for the updates!

PR #333, opened 2019-04-05
Crossterm implementation for Cursive 

TimonPost:
This PR is implementing [crossterm](https://github.com/TimonPost/crossterm) for Cursive. It is still a work in progress and has to be tested more. Down below you will find a list of thing that has to be checked or tested. 

- [X] Keyboard Input Handling
- [X] Coloring 
- [x] Mouse input Handling
    - [X] Linux
    - [x] Windows (seems to not always capture mouse)
- [x] Performance (Windows is a bit laggy sometimes)
- [x] Cursor Hiding Windows.

What is [crossterm](https://github.com/TimonPost/crossterm)? 

Crossterm is similar to termion a terminal manipulation library, however, it supports all UNIX systems and Windows systems down to Windows 7.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,7 +79,7 @@ pretty-bytes = "0.2.2"
 
 [features]
 blt-backend = ["bear-lib-terminal"]
-default = ["ncurses-backend"]
+default = ["crossterm"]
 markdown = ["pulldown-cmark"]
 ncurses-backend = ["ncurses", "maplit", "term_size"]
 pancurses-backend = ["pancurses", "maplit", "term_size"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,6 +65,7 @@ version = "0.2.0"
 optional = true
 version = "1.5.1"
 
+[dependencies.crossterm]
 optional = true
 git = "https://github.com/TimonPost/crossterm"
 branch = "fixes"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,6 +65,10 @@ version = "0.2.0"
 optional = true
 version = "1.5.1"
 
+optional = true
+git = "https://github.com/TimonPost/crossterm"
+branch = "fixes"
+
 [target.'cfg(unix)'.dependencies]
 signal-hook = "0.1.7"
 
--- a/src/backend/crossterm.rs
+++ b/src/backend/crossterm.rs
@@ -182,7 +182,7 @@ impl backend::Backend for Backend {
 
     fn screen_size(&self) -> Vec2 {
         let size = self.terminal.terminal_size();
-        return Vec2::new(size.0 as usize + 1, size.1 as usize + 1);
+        Vec2::from(size) + (1, 1)
     }
 
     fn print_at(&self, pos: Vec2, text: &str) {
@@ -196,8 +196,6 @@ impl backend::Backend for Backend {
 
             self.cursor.goto(pos.x as u16, pos.y as u16);
 
-            // as I (Timon) wrote this I figured out that calling `write_str` for unix was flushing the stdout.
-            // Current work aground is writing bytes instead of a string to the terminal.
             out.write(text.as_bytes()).unwrap();
 
             let mut dupes_left = repetitions - 1;
--- a/src/backend/crossterm.rs
+++ b/src/backend/crossterm.rs
@@ -1,0 +1,298 @@
+//! Backend using the pure-rust crossplatform crossterm library.
+//!
+//! Requires the `crossterm-backend` feature.
+
+#![cfg(feature = "crossterm")]
+
+use crate::vec::Vec2;
+use crate::{backend, theme};
+use crossterm::{
+    cursor, input, terminal, AlternateScreen, AsyncReader, Attribute,
+    ClearType, Color, Colored, InputEvent as CInputEvent,
+    KeyEvent as CKeyEvent, MouseButton as CMouseButton,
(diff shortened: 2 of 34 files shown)


Cargo.toml
```diff
@@ -74,12 +78,30 @@ pretty-bytes = "0.2.2"
 
 [features]
 blt-backend = ["bear-lib-terminal"]
-default = ["ncurses-backend"]
+default = ["crossterm"]
```
TimonPost: This was for testing, and could be changed back.

Cargo.toml
```diff
@@ -74,12 +78,30 @@ pretty-bytes = "0.2.2"
...
 termion-backend = ["termion"]
+crossterm-backend = ["crossterm"]
 unstable_scroll = []
 
 [lib]
 name = "cursive"
+
+[[example]]
+name="color"
+path="examples/colors.rs"
```
gyscos: You should already be able to do `cargo run --example color` without this in the manifest

gyscos:
Regarding performance, it seems that `write_buf` calls `<AnsiOutput as IStdout>::write`, which locks `Stdout` on each call. Maybe buffering the output could improve things a bit?

TimonPost:
Yes, it does expect XTERM, I will have a look at terminfo. 

Yes, it does lock the stdout, however, I am not sure what you mean with buffering output?. Currently, I am doing some performance enhancements for crossterm because the create has all the features it should have.

Cargo.toml
TimonPost: Did not no that, I will try that out.

gyscos:
Currently, each call to `Backend::print_at` results in a call to `Stdout::lock`.
An alternative could be to instead have it write to some lock-free structure like a simple `Vec<u8>`, and then flush everything with a single lock call per frame. Note that this is also what something like a `BufWriter` would do.

In addition, it might be interesting to write directly to `/dev/tty` rather than to `stdout` (and similarly read from `/dev/tty` rather than `stdin`). This way, an application can still interact with the terminal while leaving `stdin` and `stdout` usable for other purposes. If you do that, you'll end up with something like a `File`, which is easy to wrap in a `BufWriter` and avoid locks. This is what the termion backend does.

TimonPost:
Read is currently already reading from TTY, could you elaborate more on the frame part? Is there a function from cursive that is called each frame?

gyscos:
`Backend::refresh` should be called every frame, after drawing everything. It let's you flush the output of needed.

TimonPost:
Alright, I am working on [refactoring ](https://github.com/TimonPost/crossterm/tree/refactor) crossterm, I see a noticeable performance enhancement when removing the lock. I will be working on the refactor, and release it later this week. After that, I can upgrade the version of Cursive to that new version. Currently, as you might see in the `Cargo.tom` there is a path to crossterm, this path to the crossterm on my PC on which the refactoring is done.

gyscos:
Wow - awesome!

As a note, `Cargo.toml` also supports referencing a specific branch in a git repository, so you could potentially have:

```toml
# Cargo.toml
[dependencies.crossterm]
git = "https://github.com/TimonPost/crossterm"
branch = "refactor"
```

src/backend/crossterm.rs
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    fn has_colors(&self) -> bool {
+        // TODO: color support detection?
+        true
+    }
+
+    fn screen_size(&self) -> Vec2 {
+        let size = self.terminal.terminal_size();
+        return Vec2::new(size.0 as usize + 1, size.1 as usize + 1);
```
gyscos: This could also be `Vec2::from(size) + (1, 1)`.

TimonPost: Ok!

TimonPost:
To let you know, I think it works pretty well now, I had done some changes in crossterm because the terminal size in windows was 0-based and linux 1-based. Also, I had to adjust the windows mouse position from a mouse input event. But, I think it is ready for merging now. I don't see any noticeable problems.

gyscos:
Great, thanks!

Minor detail: do you think you could `rustfmt` everything? (Might be convenient to run this automatically from the editor on save actions).

TimonPost:
Yea, I will do that. Do you also prefer unwrapping errors or using expect?

gyscos:
* `expect` is probably better, but since we're not doing it everywhere, it's probably find to `unwrap` for now.
* Regarding the `Color`, it may be nicer to use the full path in the file: `crossterm::Color`? This way it's obvious where it's coming from.

Cargo.toml
```diff
@@ -74,12 +79,13 @@ pretty-bytes = "0.2.2"
...
 markdown = ["pulldown-cmark"]
 ncurses-backend = ["ncurses", "maplit", "term_size"]
 pancurses-backend = ["pancurses", "maplit", "term_size"]
 termion-backend = ["termion"]
+crossterm-backend = ["crossterm"]
 unstable_scroll = []
 
 [lib]
-name = "cursive"
+name = "cursive"
```
gyscos: Somehow it seems your editor is eating away the trailing newlines in this file (and in the `.gitignore`).

gyscos:
How did you run `cargo fmt`? What is your `rustfmt --version`? It seems running it, you affected most files in the repo.

TimonPost:
I rebased on master, I just ran `cargo fmt`, I can't revert the format because my commits are already squashed. 

`rustfmt --version`:

`rustfmt 1.0.3-stable (d6829d62 2019-02-14)`

TimonPost:
Closing in favor of #335

PR #335, opened 2019-04-12
Added crossterm backend.

TimonPost:
This PR is implementing [crossterm](https://github.com/TimonPost/crossterm) for Cursive. It is still a work in progress and has to be tested more. Down below you will find a list of thing that has to be checked or tested. 

- [X] Keyboard Input Handling
- [X] Coloring 
- [x] Mouse input Handling
    - [X] Linux
    - [x] Windows (seems to not always capture mouse)
- [x] Performance (Windows is a bit laggy sometimes)
- [x] Cursor Hiding Windows.

What is [crossterm](https://github.com/TimonPost/crossterm)? 

Crossterm is similar to termion a terminal manipulation library, however, it supports all UNIX systems and Windows systems down to Windows 7.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,6 +66,10 @@ version = "0.4.1"
 optional = true
 version = "1.5.1"
 
+[dependencies.crossterm]
+optional = true
+version = "0.9.2"
+
 [target.'cfg(unix)'.dependencies]
 signal-hook = "0.1.8"
 
@@ -75,11 +79,13 @@ pretty-bytes = "0.2.2"
 
 [features]
 blt-backend = ["bear-lib-terminal"]
-default = ["ncurses-backend"]
+#default = ["ncurses-backend"]
+default = ["crossterm"]
 markdown = ["pulldown-cmark"]
 ncurses-backend = ["ncurses", "maplit", "term_size"]
 pancurses-backend = ["pancurses", "maplit", "term_size"]
 termion-backend = ["termion"]
+crossterm-backend = ["crossterm"]
 unstable_scroll = []
 
 [lib]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,8 +79,7 @@ pretty-bytes = "0.2.2"
 
 [features]
 blt-backend = ["bear-lib-terminal"]
-#default = ["ncurses-backend"]
-default = ["crossterm"]
+default = ["ncurses-backend"]
 markdown = ["pulldown-cmark"]
 ncurses-backend = ["ncurses", "maplit", "term_size"]
 pancurses-backend = ["pancurses", "maplit", "term_size"]
--- a/src/backend/crossterm.rs
+++ b/src/backend/crossterm.rs
@@ -1,0 +1,302 @@
+//! Backend using the pure-rust crossplatform crossterm library.
+//!
+//! Requires the `crossterm-backend` feature.
+
+#![cfg(feature = "crossterm")]
+
+use crate::vec::Vec2;
+use crate::{backend, theme};
+use crossterm::{
+    cursor, input, terminal, AlternateScreen, AsyncReader, Attribute,
+    ClearType, Color, Colored, InputEvent as CInputEvent,
+    KeyEvent as CKeyEvent, MouseButton as CMouseButton,
+    MouseEvent as CMouseEvent, Terminal, TerminalCursor,
+};
+
+use crate::event::{Event, Key, MouseButton, MouseEvent};
+use std::cell::{Cell, RefCell};
+use std::io::{self, Stdout, Write};
+
+/// Backend using crossterm
+pub struct Backend {
+    current_style: Cell<theme::ColorPair>,
+    last_button: Option<MouseButton>,
+    // reader to read user input async.
+    async_reader: AsyncReader,
+    alternate_screen: AlternateScreen,
+    stdout: RefCell<Stdout>,
+    cursor: TerminalCursor,
+    terminal: Terminal,
+}
+
+impl Backend {
+    /// Creates a new crossterm backend.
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -74,11 +79,12 @@ pretty-bytes = "0.2.2"
 
 [features]
 blt-backend = ["bear-lib-terminal"]
-default = ["ncurses-backend"]
+default = ["crossterm"]
```
gyscos: Let's move the default back to `ncurses-backend`.

src/backend/crossterm.rs
```diff
@@ -0,0 +1,298 @@
...
+    // reader to read user input async.
+    async_reader: AsyncReader,
+    alternate_screen: AlternateScreen,
+    stdout: RefCell<Stdout>,
+    cursor: TerminalCursor,
+    terminal: Terminal,
+}
+
+impl Backend {
+    /// Creates a new dummy backend.
```
gyscos: Let's update the doc here

src/backend/crossterm.rs
```diff
@@ -0,0 +1,298 @@
...
+                            MouseEvent::Hold(self.last_button.unwrap());
+                        let position = (x - 1, y - 1).into();
+
+                        return Event::Mouse {
+                            event,
+                            position,
+                            offset: Vec2::zero(),
+                        };
+                    }
+                    _ => Event::Unknown(vec![]),
```
gyscos: We may want to add some debug logs like `log::warn!` when this happens

src/backend/crossterm.rs
```diff
@@ -0,0 +1,298 @@
...
+            }
+            _ => Event::Unknown(vec![]),
+        };
+        Event::Unknown(vec![])
+    }
+}
+
+impl backend::Backend for Backend {
+    fn poll_event(&mut self) -> Option<Event> {
+        if let Some(input_event) = self.async_reader.next() {
```
gyscos: This could be done

```rust
self.async_reader.next().map(|event| self.map_key(event))
```

src/cursive.rs
```diff
@@ -77,6 +77,13 @@ impl<F: FnOnce(&mut Cursive) -> () + Send> CbFunc for F {
     }
 }
 
+#[cfg(feature = "crossterm-backend")]
```
gyscos: You'll have to update all other `impl Default` block to add a `not(feature = "crossterm-backend")` in their `cfg` conditions, otherwise you could get overlapping implementations.

src/backend/crossterm.rs
```diff
@@ -145,22 +145,23 @@ impl Backend {
                             offset: Vec2::zero(),
                         };
                     }
-                    _ => Event::Unknown(vec![]),
+                    _ => {
+                        log::warn!("Unknown mouse button event!");
```
gyscos: Could we even log the event itself? (I think it implements `Debug`?)

src/cursive.rs
```diff
@@ -77,28 +77,38 @@ impl<F: FnOnce(&mut Cursive) -> () + Send> CbFunc for F {
     }
 }
 
-#[cfg(feature = "crossterm-backend")]
+#[cfg(all(
+not(feature = "termion-backend"),
```
gyscos: You don't have to make this one depend on `not(feature = "termion-backend")`. Otherwise, when there is both `crossterm-backend` and `termion-backend`, there will be no `Default` implementation.

In general, the conditions pattern I'm trying to get is this (in pseudo-code):

```
if A {
    Default = A
} else if B {
    Default = B
} else if C {
    Default = C
} else {
    ...
}
```

Only there is no `else` clause for `#cfg` blocks, so we have to unroll it manually, and have:

```
if A {
    Default = A
}

if B and not A:
    Default = B
}

if C and not B and not A {
    Default = C
}

...
```

src/cursive.rs
```diff
@@ -77,28 +77,38 @@ impl<F: FnOnce(&mut Cursive) -> () + Send> CbFunc for F {
...
 impl Default for Cursive {
     fn default() -> Self {
         Self::crossterm().unwrap()
     }
 }
 
-#[cfg(feature = "termion-backend")]
+#[cfg(all(
+not(feature = "crossterm-backend"),
+feature = "termion-backend")
```
gyscos: This might need some `rustfmt` here

src/cursive.rs
hcpl: You can use the [`cfg-if`](https://docs.rs/cfg-if/0.1.7/cfg_if/) crate for that:

```rust
if_cfg! {
    if #[cfg(feature = "crossterm-backend")] {
        // Default = crossterm
    } else if #[cfg(feature = "termion-backend")] {
        // Default = termion
    } else if #[cfg(feature = "pancurses-backend")] {
        // Default = pancurses
    } else if #[cfg(feature = "blt-backend")] {
        // Default = blt
    } else if #[cfg(feature = "ncurses-backend")] {
        // Default = ncurses
    }
}
```

By the way, what should be done in case if no features have been enabled? Cursive itself seems to support that case, since it builds with no default features, but `impl Default for Cursive` is missing in that case.

gyscos: Ah right, we can always defaults to the `DummyBackend` if no other built-in backends are available, though failing to compile is not such a bad option, as it often indicates a mistake somewhere (using the dummy backend through `Cursive::default()` is rarely what the user wants).

gyscos: `cfg-if` looks neat, it's just a shame it breaks `rustfmt` inside the macro block :(

gyscos:
On the latest master I started using `cfg-if` for the `impl Default for Cursive` blocks, I think that's the only conflict that needs fixing.

TimonPost:
Awesome, I think it is ready for the merge

PR #342, opened 2019-05-08
[PAN] Interprete 'carriage return' as 'enter'

zsx:
Pancurses returns different key codes for "Enter" in the normal area
('\r'), and the one in the numpad ('\n'). Cursive only interpretes '\n'
as Enter, and leaves '\r' as CtrlChar('m'), which is different than
that with NCurses on Linux, and makes 'Enter' key not trigger a clicked
event on the button.


gyscos:
Whoops, good catch! Thanks for the PR!

PR #351, opened 2019-05-25
replace CbFunc with Box<FnOnce()>

ngg:
Since Rust 1.35, Box<FnOnce()> compiles, we can remove the workaround.

This is a breaking change, version bump will be necessary, but it's more convenient to use,
closures without explicit argument types will work. (You can write `|s| s.quit` or `Cursive::quit` instead of `|s: &mut Cursive| s.quit()`)

### DIFF
--- a/examples/vpv.rs
+++ b/examples/vpv.rs
@@ -137,7 +137,7 @@ cargo run --example vpv </dev/zero >/dev/null",
             }
 
             // When we're done, shut down the application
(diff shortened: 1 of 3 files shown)


gyscos:
Hi, and thanks for the PR!
I was thinking exactly about that when reading the releases notes :)
I think I'll wait a bit more before merging it though (probably until 1.36 is stable), to give users some time to update their toolchain.

PR #360, opened 2019-07-22
Add strikethrough effect

cubetastic33:
Adds a strikethrough effect that works with termion, pancurses, and crossterm backends. For ncurses and bear_lib_terminal it does nothing.


gyscos:
Thanks for the work!

A bit unfortunate that it doesn't work on ncurses, but that's certainly better than nothin!

cubetastic33:
Sure! Ncurses doesn't seem to support strikethrough, that's why I didn't add it.

PR #361, opened 2019-07-24
Crossterm Performance Update

TimonPost:
I have been working on improving the performance of crosterm. The status can be followed here:https://github.com/TimonPost/crossterm/pull/175.  On windows, the performance increased with 60% and on Linux, the performance increased with +- 40% in a benchmark for cursor moving. 

The noticeable change is, especially in the windows implementation. Windows is way slower with terminal actions. I had the cursor movement benchmark where I did some stuff on windows it took 5 seconds to complete and on Linux it took less than one second in the new situation. 

I ran the examples for cursive, and there are still some glitches during rendering in the windows implementation however they were less bad then they were before. I think this is more a windows issue then a crossterm issue because on Linux crossterm performs without glitches. I am curious whether this is in the WinApi or the rust API.

For now, this should not be merged yet. I am still working on it, and once crossterm 0.10 releases I will update this PR as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Changelog
 
+## Next version: 0.13.0
+
+### Breaking changes
+
+- Update `enum-map` fron 0.5 to 0.6
+
+### API updates
+
+- Add `Effect::Strikethrough` (not supported on ncurses)
+- Add `ListView::remove_child`
+- Replace `xursive::CbFunc` with `Box<FnOnce>`
+- Add `ScrollView::{inner_size, is_as_{bottom, top, left, right} }`
+- Add getters for current value in `SliderView`
+- More fields made public in `cursive::logger`
+
 ## 0.12.0
 
 ### Breaking changes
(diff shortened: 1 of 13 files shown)


gyscos:
Thanks for the work!

Things I noticed (on linux at least) was clearing the screen takes a lot of time, more time than just re-writing everything with spaces.

Something I've been thinking about (but haven't had time to try yet) is buffer an entire screen, and combine writes to the same location. Right now, most views are drawn twice (once for the background + once for the actual content). And overlapping views re-draw the same area. This increases the traffic between the application and the terminal, and some terminals are pretty slow at handling this.
Spending a bit more CPU to combine the screen buffer (and avoid cursor moves) might be a solution for that.

TimonPost:
You want to have as view system calls as possible. I noticed that cursive redraws everything while most of the time this is really not needed. 

I updated the PR, and is not ready to me merged

gyscos:
> I updated the PR, and is not ready to me merged

Do you mean it's now ready to be merged? Or should I wait for other commits?

TimonPost:
haha, sorry made a mistake there. It is ready indeed!

gyscos:
Thanks!

PR #364, opened 2019-07-30
Update docs, fix minor typos, rephrase some parts of the tutorials

cubetastic33:
Also mention that strikethrough doesn't work with ncurses and blt.


gyscos:
Thanks a lot!

PR #366, opened 2019-08-01
Fix out-of-bounds access on TextView

leonardolang:
Frequent updates to `TextContent` while the screen is being refreshed can trigger an out-of-bounds access: an update with a smaller string being performed between calls to `compute_rows()` and `draw()` on TextView will cause an out-of-bounds access when slicing the string at `Segment::resolve`.

This change fixes the issue by caching the content value when performing size calculations and using this cached content when drawing the view.

Reference stacktrace: [textview-stacktrace.txt](https://github.com/gyscos/cursive/files/3454900/textview-stacktrace.txt)

### DIFF
--- a/src/views/text_view.rs
+++ b/src/views/text_view.rs
@@ -1,4 +1,6 @@
+use std::cell::RefCell;
 use std::ops::Deref;
+use std::ptr;
 use std::sync::Arc;
 use std::sync::{Mutex, MutexGuard};
 
@@ -11,6 +13,9 @@ use crate::utils::lines::spans::{LinesIterator, Row};
 use crate::utils::markup::StyledString;
 use crate::view::{SizeCache, View};
 use crate::{Printer, Vec2, With, XY};
+
+// Content type used internally for caching and storage
+type InnerContentType = RefCell<Arc<StyledString>>;
 
 /// Provides access to the content of a [`TextView`].
 ///
@@ -42,11 +47,12 @@ impl TextContent {
     where
         S: Into<StyledString>,
     {
-        let content = content.into();
+        let content = Arc::new(content.into());
 
         TextContent {
             content: Arc::new(Mutex::new(TextContentInner {
-                content,
+                content_value: RefCell::new(content),
+                content_cache: RefCell::new(Arc::new(StyledString::default())),
                 size_cache: None,
             })),
         }
@@ -60,18 +66,22 @@ impl TextContent {
 /// [`StyledString`]: ../utils/markup/type.StyledString.html
 ///
 /// This keeps the content locked. Do not store this!
+
 pub struct TextContentRef {
-    handle: OwningHandle<
+    _handle: OwningHandle<
         ArcRef<Mutex<TextContentInner>>,
         MutexGuard<'static, TextContentInner>,
     >,
+    // We also need to keep a copy of Arc so `deref` can return
+    // a reference to the `StyledString`
+    data: Arc<StyledString>,
 }
 
 impl Deref for TextContentRef {
     type Target = StyledString;
 
     fn deref(&self) -> &StyledString {
-        &self.handle.content
+        self.data.as_ref()
     }
 }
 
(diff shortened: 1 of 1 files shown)


gyscos:
Arf, right! Thanks for noticing, and for providing a fix!!

I only wish we could have a more efficient solution, at least in the case where no asynchronous update happens - `compute_rows` can be called quite often (for example when resizing the terminal).
I wonder if we could either:
* Avoid invalid rows by re-computing the rows in `draw` if `TextContent::size_cache` is gone. Con: `draw` takes a `&mut self`, so we'd need to add some more interior mutability.
* Only create the clone when actually invalidating the data, in `TextContent::with_content` (and only the first time since `compute_rows` was called).

I'll have a look at this when I get time.

leonardolang:
No worries, glad to help!

Indeed, I did not consider the frequent resize case (or a string with considerable size).

After putting some thought into optimizing the second approach I came up with a solution which uses `RefCell<Arc<StyledString>>` on `TextContentInner`, sharing same underlying `Arc` for the data when cache is valid - and only performing a clone with a frequency of `min(refresh_rate, update_rate)`.

Initially, I implemented this approach using `RefCell<Rc<StyledString>>` and tests actually passed - but when testing on my scenario it failed to compile since I'm using threads. I was looking for a way around the double `Arc` as well, but even moving the cache to the `TextView` would still require an additional `Arc` on `TextContent::content` for allowing sharing of the `String`, so there's not much gain there.

In any case, I updated the PR with the new code, let me know if this approach looks interesting.

src/views/text_view.rs
```diff
@@ -12,6 +14,9 @@ use crate::utils::markup::StyledString;
 use crate::view::{SizeCache, View};
 use crate::{Printer, Vec2, With, XY};
 
+// Content type used internally for caching and storage
+type InnerContentType = RefCell<Arc<StyledString>>;
```
gyscos: Do you need a `RefCell` here? `TextContentInner` is behind a `Mutex`, so anytime it is accessed it should mutable; we probably don't need this extra interior mutability.

In addition, `RefCell` is `!Sync` so even with `Arc`, it probably won't work well with threads.
Note: we probably need some test to assert than `TextContent` is `Send`.

src/views/text_view.rs
```diff
@@ -81,15 +91,27 @@ impl TextContent {
...
+            // This will only clone content if content_cached and content_value
+            // are sharing the same underlying Rc.
+            if c.is_content_shared() {
+                c.content_value
+                    .replace(Arc::new(c.get_value().as_ref().clone()));
+            };
+            Arc::get_mut(&mut c.content_value.borrow_mut())
+                .expect("should not have a shared content here")
+                .append(content)
+        })
```
gyscos: Maybe you want `Arc::make_mut` here?

src/views/text_view.rs
```diff
@@ -146,6 +171,18 @@ impl TextContentInner {
             Some(ref last) => last.x.accept(size.x) && last.y.accept(size.y),
         }
     }
+
+    fn get_value(&self) -> Arc<StyledString> {
```
gyscos: Let's return a `&mut Arc` instead, and let users clone it if needed.
This will enable the `Arc::make_mut` in `append()` above.

src/views/text_view.rs
```diff
@@ -146,6 +171,18 @@ impl TextContentInner {
...
+        Arc::clone(&*self.content_value.borrow())
+    }
+
+    fn get_cache(&self) -> Arc<StyledString> {
```
gyscos: Here too let's return `&mut Arc` - most of the time we won't even need to clone the `Arc`.

src/views/text_view.rs
```diff
@@ -318,31 +352,29 @@ impl TextView {
...
         if content.is_cache_valid(size) {
             return;
         }
 
         // Completely bust the cache
         // Just in case we fail, we don't want to leave a bad cache.
         content.size_cache = None;
+        content
+            .content_cache
+            .replace(content.content_value.borrow().clone());
```
gyscos: Let's use `Arc::clone(...)` rather than `.clone()` to be clear about what we're cloning here.

gyscos:
Thanks for the updated solution! Just a few details and it should be good to go! :)

gyscos:
I ended up merging it and adding the fixes on top of that - thanks again!

leonardolang:
Hi there, sorry for not responding before - I was planning to look at this when I got some free time but then I got caught up with a few other things. Thanks for the awesome library btw. :)

PR #368, opened 2019-08-17
BoxView Fixed, AtLeast, AtMost SizeConstraints

jwuensche:
This PR contains a fix for `BoxView`, currently there is a bug in the behaviour of the view, which makes it display unexpectetly for the user.
Namingly that a `BoxView` with `SizeConstraints` of `Fixed`, `AtLeast` & `AtMost` can be rendered outside of these constraints, when they are enclosed in a `ViewWrapper` like another `BoxView`, or a `LinearLayout`.

In the current state the BoxView is just the maximum size, as can be reproduced in this minimal example:

```rust
use cursive::{Cursive, views::{BoxView, TextView, Panel}};

fn main() {
    let mut siv = Cursive::default();

    siv.add_layer(
        Panel::new(
            BoxView::with_fixed_size(
                (40,20),
                BoxView::with_fixed_size(
                    (20,10),
                    // This is borked, when you run the example you can see that the panel
                    // is just the complete inner region of the outer panel while it should
                    // be half its height and width
                    Panel::new(
                        TextView::new("I'm 20 x 10 !")
                    )
                )
            )
        )
    );
    siv.run();
}
```

This results in:

```
                   ┌────────────────────────────────────────┐
                   │┌──────────────────────────────────────┐│
                   ││I'm 20 x 10 !                         ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   │└──────────────────────────────────────┘│
                   └────────────────────────────────────────┘
```

Which is not would have been expected when we defined a **fixed** size BoxView.

The problem is that currently the `SizeConstraint` of `Fixed` is not correctly handled in the ViewWrapper implementation of the BoxView.

Also problematic is when a Box is set to be a `Fixed` or `AtLeast` size and the current printer is smaller than this, we would be expect that this `BoxView` is then cutoff to not violate the restriction of Box, but the current result is:

```
                   ┌────────────────────────────────────────┐
                   │┌──────────────────────────────────────┐│
                   ││I'm 50 x 30 !                         ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   ││                                      ││
                   │└──────────────────────────────────────┘│
                   └────────────────────────────────────────┘

```

Which is then exactly the same as before, even if the `SizeConstraint` is normally out of range in this scenario.

The commit in this PR contains my proposal to solve this problem. It produces for the same scenarios the outputs:

```
                   ┌────────────────────────────────────────┐
                   │┌───────────────────────────────────────│
                   ││I'm 50 x 30 !                          │
                   ││                                       │
                   ││                                       │
                   ││                                       │
                   ││                                       │
                   ││                                       │
                   ││                                       │
                   ││                                       │
                   ││                                       │
                   ││                                       │
                   ││                                       │
                   ││                                       │
                   ││                                       │
                   ││                                       │
                   ││                                       │
                   ││                                       │
                   ││                                       │
                   ││                                       │
                   ││                                       │
                   └────────────────────────────────────────┘

```

and

```
                   ┌────────────────────────────────────────┐
                   │┌──────────────────┐                    │
                   ││I'm 20 x 10 !     │                    │
                   ││                  │                    │
                   ││                  │                    │
                   ││                  │                    │
                   ││                  │                    │
                   ││                  │                    │
                   ││                  │                    │
                   ││                  │                    │
                   │└──────────────────┘                    │
                   │                                        │
                   │                                        │
                   │                                        │
                   │                                        │
                   │                                        │
                   │                                        │
                   │                                        │
                   │                                        │
                   │                                        │
                   │                                        │
                   └────────────────────────────────────────┘

```

### DIFF
--- a/src/views/box_view.rs
+++ b/src/views/box_view.rs
@@ -1,3 +1,4 @@
+use crate::printer::Printer;
 use crate::vec::Vec2;
 use crate::view::{SizeConstraint, View, ViewWrapper};
 use crate::With;
@@ -212,12 +213,25 @@ impl<T: View> BoxView<T> {
 impl<T: View> ViewWrapper for BoxView<T> {
     wrap_impl!(self.view: T);
 
+    fn wrap_draw(&self, printer: &Printer) {
+        let mut printer = printer.clone();
+        // Set values hard here, since cropping would allow smaller size if fixed or at least set
+        printer.size.x = self.size.x.result((printer.size.x, printer.size.x));
+        printer.size.y = self.size.y.result((printer.size.y, printer.size.y));
+        self.view.draw(&printer);
+        self.view.draw(&printer);
+    }
+
     fn wrap_required_size(&mut self, req: Vec2) -> Vec2 {
         // This is what the child will see as request.
         let req = self.size.zip_map(req, SizeConstraint::available);
 
         // This is the size the child would like to have.
-        let child_size = self.view.required_size(req);
+        // Given the constraints of our box.
+        let child_size = self.view.required_size(Vec2::new(
+            self.size.x.result((req.x, req.x)),
+            self.size.y.result((req.y, req.y)),
+        ));
 
         // Some of this request will be granted, but maybe not all.
         let result = self
--- a/src/views/box_view.rs
+++ b/src/views/box_view.rs
@@ -214,12 +214,9 @@ impl<T: View> ViewWrapper for BoxView<T> {
     wrap_impl!(self.view: T);
 
     fn wrap_draw(&self, printer: &Printer) {
-        let mut printer = printer.clone();
-        // Set values hard here, since cropping would allow smaller size if fixed or at least set
-        printer.size.x = self.size.x.result((printer.size.x, printer.size.x));
-        printer.size.y = self.size.y.result((printer.size.y, printer.size.y));
-        self.view.draw(&printer);
-        self.view.draw(&printer);
+        self.view.draw(&printer.inner_size(
+            self.size.zip_map(printer.size, |c, s| c.result((s, s))),
+        ));
     }
 
     fn wrap_required_size(&mut self, req: Vec2) -> Vec2 {
@@ -228,10 +225,7 @@ impl<T: View> ViewWrapper for BoxView<T> {
 
         // This is the size the child would like to have.
         // Given the constraints of our box.
-        let child_size = self.view.required_size(Vec2::new(
-            self.size.x.result((req.x, req.x)),
-            self.size.y.result((req.y, req.y)),
-        ));
+        let child_size = self.size.zip_map(req, |c, s| c.result((s, s)));
 
         // Some of this request will be granted, but maybe not all.
         let result = self


src/views/box_view.rs
```diff
@@ -212,12 +213,25 @@ impl<T: View> BoxView<T> {
 impl<T: View> ViewWrapper for BoxView<T> {
     wrap_impl!(self.view: T);
 
+    fn wrap_draw(&self, printer: &Printer) {
+        let mut printer = printer.clone();
+        // Set values hard here, since cropping would allow smaller size if fixed or at least set
+        printer.size.x = self.size.x.result((printer.size.x, printer.size.x));
+        printer.size.y = self.size.y.result((printer.size.y, printer.size.y));
+        self.view.draw(&printer);
+        self.view.draw(&printer);
```
gyscos: Duplicated line?

src/views/box_view.rs
```diff
@@ -212,12 +213,25 @@ impl<T: View> BoxView<T> {
 impl<T: View> ViewWrapper for BoxView<T> {
     wrap_impl!(self.view: T);
 
+    fn wrap_draw(&self, printer: &Printer) {
+        let mut printer = printer.clone();
+        // Set values hard here, since cropping would allow smaller size if fixed or at least set
+        printer.size.x = self.size.x.result((printer.size.x, printer.size.x));
+        printer.size.y = self.size.y.result((printer.size.y, printer.size.y));
```
gyscos: You could use `Printer::cropped` and `Vec2::zip_map`:
```rust
self.view.draw(&printer.cropped(
    self.size.zip_map(printer.size, |c, s| c.result((s, s)))
));
```

src/views/box_view.rs
```diff
@@ -212,12 +213,25 @@ impl<T: View> BoxView<T> {
...
     fn wrap_required_size(&mut self, req: Vec2) -> Vec2 {
         // This is what the child will see as request.
         let req = self.size.zip_map(req, SizeConstraint::available);
 
         // This is the size the child would like to have.
-        let child_size = self.view.required_size(req);
+        // Given the constraints of our box.
+        let child_size = self.view.required_size(Vec2::new(
+            self.size.x.result((req.x, req.x)),
+            self.size.y.result((req.y, req.y)),
```
gyscos: Here too you could so `self.size.zip_map(req, |c, r| c.result((r, r)))`.

gyscos:
Wow, thanks for the analysis, and for the PR!
Just a few minor nits. Will try it a bit and it should be good to go!

src/views/box_view.rs
jwuensche: Oh yeah, thanks for the hint, overlooked this while rebasing

src/views/box_view.rs
jwuensche: Good suggestion, used the `zip_map` in the new commit. But using the `cropped` of the printer doesn't work at the moment e.g. the PanelView as I used in the example still prints the full panel with the size of the printer. But `inner_size` is applicable for in this situation.

src/views/box_view.rs
jwuensche: Changed, to this. Again thanks for the suggestion :+1:

src/views/box_view.rs
gyscos: Ah right sorry - you'll want `printer.inner_size()` in this case, since we're only changing the size visible to the child.

gyscos:
Thanks again!

PR #374, opened 2019-09-01
Fix clippy warning in tutorial 3

mimame:
- Writing `&String` instead of `&str` involves a new object where a  slice will do
  See https://rust-lang.github.io/rust-clippy/master/#ptr_arg


gyscos:
Thanks! I thought I remember this not compiling (when giving it to a parameter expecting a `Fn(&T)` with `T = String`) but it seems to work! :tada:

PR #376, opened 2019-09-07
Change `TitleSecondary` from Yellow to Light Blue

jwuensche:
Hi,
I've recently played around with themes in cursive and noticed that in many commonly used themes the `TitleSecondary` color currently used is hard to distinguish from either `TitlePrimary` or `View` colors.

The current palette defaults sets `TitleSecondary` to a dark yellow:

```
Background => Dark(Blue),
Shadow => Dark(Black),
View => Dark(White),
Primary => Dark(Black),
Secondary => Dark(Blue),
Tertiary => Dark(White),
TitlePrimary => Dark(Red),
TitleSecondary => Dark(Yellow),
Highlight => Dark(Red),
HighlightInactive => Dark(Blue),
```
This results in looks similar to these:
#####  XTerm
xterm_not_bright
Making it hard to read the complete text at first glance because the bright color of the text on a bright background.
##### Rxvt
rxvt

Same for Rxvt.

##### Tango
tango_not_bright

The variance between both is not that large. If we enable now bold text as bright text the behaviour is not that different, we still might have issues differentiating text from background without a closer look.

##### XTerm
XTerm
 
##### Rxvt
Rxvt

##### Tango
Tango

For this PR I have changed the used color in the default palette to a light blue, for one to increase the difference of color to  `TitlePrimary` and `View` and to use a color otherwise not occupied in the current default palette.

##### XTerm
blue_xterm_non

##### Rxvt
blue_rxvt_no

##### Tango
blue_tango_non

And with bold as bright text active:
##### XTerm
blue_xterm

##### Rxvt
blue_urxvt

##### Tango
blue_tango

### DIFF
--- a/src/theme/palette.rs
+++ b/src/theme/palette.rs
@@ -161,7 +161,7 @@ impl Default for Palette {
                 Secondary => Dark(Blue),
                 Tertiary => Dark(White),
                 TitlePrimary => Dark(Red),
-                TitleSecondary => Dark(Yellow),
+                TitleSecondary => Light(Blue),
                 Highlight => Dark(Red),
                 HighlightInactive => Dark(Blue),
             },


gyscos:
Indeed, much more readable! Thanks!

PR #385, opened 2019-10-04
Allow SelectView::with_all to accept StyledString

panicbit:
Changes the bound on the iterator items from `Into<String>` to `Into<StyledString>`.

This brings the API in line with the other item methods,
which all use the same bound.

---
This change is [Reviewable](https://reviewable.io/reviews/gyscos/cursive/385)


gyscos:
Thanks for the PR!

PR #390, opened 2019-10-10
Don't use StackView::draw_bg on Cursive::draw

Riey:
`StackView::draw_bg` is too expensive because it's call `Backend::print_at` many times(X * Y)

It could be simple with just one call of `Backend::clear`

### DIFF
--- a/src/cursive.rs
+++ b/src/cursive.rs
@@ -36,10 +36,6 @@ pub struct Cursive {
     global_callbacks: HashMap<Event, Vec<Callback>>,
     menubar: views::Menubar,
 
-    // Last layer sizes of the stack view.
-    // If it changed, clear the screen.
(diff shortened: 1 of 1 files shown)


gyscos:
Thanks for the PR!

Actually we used to `clear()` each frame like that, but it turns out that the clear instruction is very expensive on most terminals; in many cases, printing empty lines is faster.

Note that `StackView::draw_bg` doesn't actually call `Backend::print_at` x*y times, it uses `Printer::print_hline` Y times, which is usually implemented in a smarter way and doesn't insert "move" commands everywhere.

Did you notice `StackView::bg` being too slow? In general performance-driven changes like that may benefit from some measurements.

Riey:
Sorry I didn't know `clear` is far more expensive

PR #391, opened 2019-10-10
Change lazy_static version specifier to 1

dbrgn:
This allows cursive to be used together with libraries that pin the
`lazy_static` dependency to a version below 1.3.


gyscos:
Thanks for the PR!
At some point I may want to compute the actual minimum required version and have better bounds...

dbrgn:
Thanks! Any chance for a release?

PR #395, opened 2019-10-25
Make enter and tab key work again in crossterm backend

### LINKED ISSUE #394: Enter key doesn't work in crossterm backend
When compiling the examples, for example `progress`  with commit dac0f5a67e8d0dce5faa2968cf96de327998a660) with the crossterm backend:
```
cargo build --examples --no-default-features --features crossterm-backend
```
The enter key to select a menu option doesn't work. This broke some time recently, I think when going from crossterm 0.10 to 0.12,
 
It's working with commit 565c64c528e26f7a5bb3349e85a2f1b3d77a8ee7. Bisecting is a problem, most of the intermediate commits don't build due to a crossterm build issue.

vmedea:
It looks like new crossterm returns a dedicated enum item for the enter and tab key instead of the `\n` and `\t` character.

Fixes #394.


gyscos:
Whoops, thanks for noticing!

PR #399, opened 2019-10-27
crossterm: Correct attribute for end of underline and reverse effect

vmedea:
- `NoUnderline` instead of `Underline`
- `NoInverse` instead of `Reverse`


gyscos:
Thanks!

PR #400, opened 2019-10-27
crossterm: Stop mouse mode and show cursor at finish

vmedea:
Something is wrong at the moment when finishing up, when the crossterm backend is used.

The terminal is left in a strange state: the previous contents of the terminal (as before the program) are visible, however the cursor is invisible (and reset to the top left), and clicking the mouse spams
escape codes.

For some reason the `execute!` block in `finish` is failing. But it doesn't seem to be necessary: crossterm already restores the alternative screen at shutdown, which also restores the original cursor position and attributes.

Do make sure that mouse mode is stopped and that the cursor is visible.

### DIFF
--- a/src/backend/crossterm.rs
+++ b/src/backend/crossterm.rs
@@ -7,10 +7,10 @@
 use crate::vec::Vec2;
 use crate::{backend, theme};
 use crossterm::{
-    cursor, execute, input, queue, terminal, AlternateScreen, AsyncReader,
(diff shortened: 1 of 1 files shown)


gyscos:
Ah yes - the crossterm backend is still a bit fresh, and not quite as well-tested as the others. Thanks for the work!

vmedea:
Yes, it's somewhat finnicky at the moment. Still it's surprising how well it works, given how simple the backend is.

Issue #401, opened 2019-10-28
Inconsistent rendering in `markup` example in crossterm and termion


vmedea:
In some terminals, the `markup` example is rendered inconsistently for termion and crossterm backend. This doesn't happen for the ncurses backend.

in tmux (after pressing `Tab` once)
![tmux](https://user-images.githubusercontent.com/43048142/67663200-3eac4080-f965-11e9-8ef2-a2e2c4a80bd8.png)

in gnome-terminal:
![gnome-terminal](https://user-images.githubusercontent.com/43048142/67663212-453ab800-f965-11e9-8aff-1d12e89147b8.png)

Also reported here: https://github.com/gyscos/cursive/issues/90#issuecomment-515743068

The root problem is that the `NoBold` (21) terminal attribute doesn't reliably work over terminals https://github.com/crossterm-rs/crossterm/issues/294

It appears that `NormalIntensity` (22) instead works more portably to cancel bold. ~~That catch is that this ends italic as well, so this would need some more logic… (this doesn't seem to be true)~~

gyscos:
Blast! :^/

I guess one way would be to reset all the attributes, and re-enable the ones we want to keep. The backend code would have to store the current style; when calling `unset_effect`, it would update this current style, send a big `Reset` that disables all attributes, then re-set the current style. It means a few more escape codes to send, but it shouldn't be _that_ bad most of the time (except when assigning a different style to each cell in a nethack-style where it would ~double the required bandwidth).

I'm curious how `termion` and `ncurses` handle this - do they handle this state internally? Or are they using a different escape code? Will need to investigate.

vmedea:
> I guess one way would be to reset all the attributes, and re-enable the ones we want to keep.

Resetting all attributes every time a state is disabled would be the surest, but that'd involve re-sending the foreground and background color too!

Good news: I've done some testing on the affected terminals and it seems that using 22 instead of 21 "just works". It doesn't disable italic or any of the other supported attributes, as I thought.
```bash
echo -e "\x1b[1m bold enabled\x1b[22m disabled"
echo -e "\x1b[3m italic\x1b[22m (still italic)\x1b[23m disabled"
echo -e "\x1b[4m underline\x1b[22m (still underline)\x1b[24m disabled"
echo -e "\x1b[7m reverse\x1b[22m (still reverse)\x1b[27m disabled"
echo -e "\x1b[9m strikethrough\x1b[22m (still strikethrough)\x1b[29m disabled"
```
[...]

PR #402, opened 2019-10-28
crossterm, termion: Reset bold with SGR 22 instead of 21

vmedea:
NoBold (21) doesn't work in many terminals to disable bold.
22 (named NormalIntensity in crossterm, and NoFaint in termion) seems to do a better job.

Attempt at fixing #401.


gyscos:
Thanks a lot for the work! Seems to be working fine here (tried the linux terminal, gnome-terminal, urxvt, alacritty and konsole).

PR #403, opened 2019-11-03
Fix crossterm mouse event position

Riey:
Fix crossterm backend give correct mouse position

## Previous example
menubar example

**About** button is clicked when click one line below

Screenshot_20191103_194136


gyscos:
Thanks for the work!!

vmedea:
ha, yes, crossterm coordinates already start from 0

Issue #352, opened 2019-05-26
Crossterm panics when setting theme


chrisvest:
# Calling set_theme with the crossterm backend hits a code path that is unimplemented.

Specifically, this place: https://github.com/gyscos/Cursive/blob/master/src/backend/crossterm.rs#L293

I don't know if this was forgotten about in the original PR or intentionally punted. XV always sets a theme when it starts, so it always panics on startup when built with the crossterm backend.

```
Panic in xv 0.1.1, 2019-05-26T16:14:19+02:00.
System: linux (unix), x86_64.
Location: /home/chris/.cargo/registry/src/github.com-1ecc6299db9ec823/cursive-0.12.0/src/backend/crossterm.rs:293:13.

stack backtrace:
   0: backtrace::backtrace::libunwind::trace::h8ddc7da5b84baa88 (0x560d2585d04c)
             at /home/chris/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.16/src/backtrace/libunwind.rs:80
      backtrace::backtrace::trace_unsynchronized::h382fc165cd29745b
             at /home/chris/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.16/src/backtrace/mod.rs:55
   1: backtrace::backtrace::trace::hf72464fdbee526f8 (0x560d2585cfcd)
             at /home/chris/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.16/src/backtrace/mod.rs:46
   2: backtrace::capture::Backtrace::create::h720ed5cf10bf8045 (0x560d25854684)
             at /home/chris/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.16/src/capture.rs:132
   3: backtrace::capture::Backtrace::new::h355a6f3c77f26977 (0x560d25854573)
             at /home/chris/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.16/src/capture.rs:97
   4: xv::panic_hook::report_crash::h77e3b591e63723b0 (0x560d2583279e)
             at src/panic_hook.rs:58
   5: xv::panic_hook::install::{{closure}}::h7a3fe381ae0cd7ae (0x560d257772fa)
             at src/panic_hook.rs:19
   6: rust_panic_with_hook (0x560d259ac288)
             at src/libstd/panicking.rs:478
   7: continue_panic_fmt (0x560d259abd21)
[...]

PR #406, opened 2019-11-11
Crossterm 0.14 

TimonPost:
A lot of changes have happened since 0.12, going to skip 0.13 and aim for the coming 0.14 in this PR. Which is going to have a very nice event reading system.  People can use this branch if they want to be on the latest crossterm, however, the dependency branch is still a WIP, and changes can still happen and it might be broken sometimes. We do have settled on the public API. So I don't think anything is going to be changed about that. 

- Implemented back/foreground color support
- Migrated 0.13 changes
- Migrated 0.14 changes 
- Removed `AsyncReader`
- Reordered trait methods
- Implemented terminal reseize events
- Cleaned up some code
- Moved parsing logic into `From` trait implementations.


src/backend/crossterm.rs
```diff
@@ -169,16 +165,22 @@ impl backend::Backend for Backend {
     }
 
     fn poll_event(&mut self) -> Option<Event> {
-        self.async_reader.next().map(|event| self.map_key(event))
+        match poll(Some(Duration::from_millis(10))) {
```
TimonPost: Here I poll for 10 milliseconds. If there is no input for 10 millis it will return `Ok(false)`

gyscos: Do we need to poll at all? We already have some wait/polling in place outside of the backend, would it work if it was entirely non-blocking?

examples/menubar.rs
```diff
@@ -9,8 +9,8 @@ use std::sync::atomic::{AtomicUsize, Ordering};
 // application.
 
 fn main() {
+    println!("a");
```
gyscos: Probably don't this need.

src/backend/crossterm.rs
TimonPost: We can make it a duration of 0 millis. This will almost immediately return the poll result.

TimonPost:
Also, we are going to support resize events in this version. Which might be very interesting for this lib.

TimonPost:
The PR for input rewrite has been merged. We can expect a release in one week.

TimonPost:
I did an update, you might want to review again. Especially the place were we map events.  Crossterm stored modifier information separate from the event. Cursive does it as crossterm did before, this requires some mapping, and hope that all of this went well. It looks like it tough. 

I haven't jet tested the mouse events.

src/backend/crossterm.rs
```diff
@@ -4,28 +4,140 @@
...
+                code: KeyCode::Char(c),
+            } => Event::CtrlChar(c),
+            CKeyEvent {
+                modifiers: KeyModifiers::ALT,
+                code: KeyCode::Char(c),
+            } => Event::AltChar(c),
+            CKeyEvent {
+                modifiers: _,
+                code: KeyCode::Char('c'),
+            } => Event::Exit,
```
gyscos: Maybe we want exit for `Ctrl-C`, not for just `'c'`?

src/backend/crossterm.rs
```diff
@@ -4,28 +4,140 @@
...
+            CKeyEvent {
+                modifiers: KeyModifiers::CONTROL,
+                code,
+            } => Event::Ctrl(Key::from(code)),
+            CKeyEvent {
+                modifiers: KeyModifiers::ALT,
+                code,
+            } => Event::Alt(Key::from(code)),
+            CKeyEvent {
+                modifiers: KeyModifiers::SHIFT,
```
gyscos: When writing uppercase letters, will the modifier be `SHIFT`?

TimonPost: No, when you enabled capslock, then no SHIFT is sent. Only an  uppercase character.

src/backend/crossterm.rs
TimonPost: So a capital C?

gyscos: I mean, right now, doesn't this mean that pressing the letter <kbd>c</kbd> will close the program? Is that desired?

src/backend/crossterm.rs
gyscos: What about pressing <kbd>Shift</kbd>+<kbd>A</kbd> to send a single uppercase `'A'`?

TimonPost: That is correct, I was not sure how to handle this case. Because cursive doesn't seem to have the `ShiftChar()` field. How is this handled currently? I would like to support SHIFT + CHAR as well.

gyscos: In cursive it's just a `Event::Char` with the uppercase character, like `Event::Char('A')` vs `Event::Char('a')`.

TimonPost: Ah, okay, I can add that behaviour

src/backend/crossterm.rs
TimonPost: Correct, I looked at [this](https://github.com/gyscos/cursive/blob/master/src/backend/termion.rs#L123) it also has Ctrl + c. I will do an update

src/backend/crossterm.rs
```diff
@@ -36,149 +155,96 @@ impl Backend {
...
+                Err(_) => None,
+            },
+            _ => None,
+        }
     }
 
     fn finish(&mut self) {
-        input().disable_mouse_mode().unwrap();
-        cursor().show().unwrap();
+        execute!(self.stdout_mut(), DisableMouseCapture, Show)
```
gyscos: Currently exiting an application crashes here: 
```
thread 'main' panicked at 'already borrowed: BorrowMutError', src/libcore/result.rs:1187:5
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:84
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:61
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1030
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1412
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:65
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:50
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:188
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:205
  10: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:464
  11: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:373
  12: rust_begin_unwind
             at src/libstd/panicking.rs:302
  13: core::panicking::panic_fmt
             at src/libcore/panicking.rs:82
  14: core::result::unwrap_failed
             at src/libcore/result.rs:1187
  15: core::result::Result<T,E>::expect
             at /rustc/5c5b8afd80e6fa1d24632153cb2257c686041d41/src/libcore/result.rs:982
  16: core::cell::RefCell<T>::borrow_mut
             at /rustc/5c5b8afd80e6fa1d24632153cb2257c686041d41/src/libcore/cell.rs:876
  17: cursive::backend::crossterm::Backend::stdout_mut
             at src/backend/crossterm.rs:180
  18: <cursive::backend::crossterm::Backend as cursive::backend::Backend>::finish::{{closure}}
             at src/backend/crossterm.rs:242
  19: core::result::Result<T,E>::and_then
             at /rustc/5c5b8afd80e6fa1d24632153cb2257c686041d41/src/libcore/result.rs:713
  20: <cursive::backend::crossterm::Backend as cursive::backend::Backend>::finish
             at src/backend/crossterm.rs:242
```

gyscos:
Thanks a lot for all the work!

Key input seems to be working well!

Mouse input is weird though: it seems there's a delay between each event, maybe 300-500ms, which means that things like dragging the mouse, which creates a lot of events, cause significant input lag.

src/backend/crossterm.rs
gyscos: Note that since we have `&mut self` here, we can directly use `RefCell::get_mut`:
```suggestion
        execute!(self.stdout.get_mut(), DisableMouseCapture, Show)
```

TimonPost: That is strange. Should not happen. I haven't yet debugged on Linux btw.

TimonPost:
We might have to change the way we're reading the terminal size. This is done each time a mouse event occurs. We changed it to use TPUT, however, this is a separate process that crossterm spawns. This might cause the delay.

src/backend/crossterm.rs
gyscos: Found the problem: in `write_cout!`, you use `$write` twice:
https://github.com/crossterm-rs/crossterm/blob/master/src/utils/macros.rs#L20

The macro expands `$write` to the _expression_ `self.stdout_mut()`, not to the result. So it ends up calling `stdout_mut()` twice in there, which obviously fails.

TimonPost:
I noticed the same now, I will take a look into this delay, and report back.

src/backend/crossterm.rs
TimonPost: I did a macro rewrite in this PR, does switching to this branch fixes this? 

https://github.com/crossterm-rs/crossterm/pull/326

TimonPost:
I monitored the events crossterm sents. 

Mouse events: with slow-normal dragging speed about 20-26 a second.
Resize events: with slow-normal resizing speed about 10-15 a second
Keyboard events: Average typing speed: 7-10 a second. 

Mouse events are fired much more often than other events. This might explain the delay. I think cursive doesn't handle events that fast, can this be correct? Do you know how many times a second cursive polls for events? 

We could filter mouse events by skipping 3 events or something like that. We lose information, but the question is do we need te fine-grained location updates.

TimonPost:
I found the fix, it is interesting tough. I can increase the poll counter to something like: "20ms" and then it will work smoothly. 

To figure out what the reason might be I have to ask the following questions:
* Does cursive poll for events with a separate thread? 
* How often does cursive poll for events?

FreeFull:
As far as I know, cursive is entirely single-threaded. Polling for events is done in a loop in `Cursive::process_events`, in `cursive.rs`, so all of the events should be consumed before the delay kicks in. The delay is 30ms, so having a large latency is mysterious, unless there is some sort of inefficiency is pegging the CPU core at 100%, I don't know what the cause could be.

gyscos:
Note that cursive can process key events much faster: for example, copy/paste some text, it will be sent as individual key events, and you'll see that large blocks of text can be handled nearly instantly.

Cursive _can_ be thrown off if, even though there are still events pending, polling at some point returns `None`. This will cause cursive to wait for ~30ms before handling the next event. I suspect that's what's happening. `None` should only be returned if no event is available for polling; if "something unknown" is available, you can use `Event::Refresh` or `Event::Unknown`.

TimonPost:
I made a comment in our discord about it, we will have to look into this. I wrote an example for crossterm replicating the same behavior. So this has nothing to do with cursive in the way that there is an bug or something. Cursive might have influence into this problem tough.

_showcase without dealy_
```rust
fn print_events_1() -> Result<()>{
    loop {
        if poll(Duration::from_millis(0))? {
            let event = read()?;

            if let Event::Mouse(m) = event {
                println!("{:?}\r", event);
            }
        } else {
            println!("NONE\r")
        }
    }
}
```

When I run this example, my mouse events are nearly coming trough, and NONE is printed very often.

- Maybe our reader is returning way to fast. 
- Maybe this is correct, behavior, and the user should just reduce the amount of polling.
- Try to set `5ms` instead of `0ms` delay, you will notice that the return results will improve dramatically. Therefore my first statement might make sense.

I will post something, as soon, I know the answer. For now, the `5ms` delay seems to work pretty fine.

TimonPost:
We found some problems in our implementation, [this PR](https://github.com/crossterm-rs/crossterm/pull/333) has a work in progress solution.

gyscos:
Thank you very much for the investigation and the work on a solution! :)

zrzka:
## Some numbers

Couple of benchmarks ... I've got `test.txt` file (10MB, to be more precise `10_836_507` bytes).

### ViM and copy & paste

When I paste the content into an empty buffer, it takes 3m50s to finish the operation. To be fair, it's pretty _useless_ as ViM prints all these characters, but the following examples do not.

### crossterm and copy & paste

* [cargo run --release --example perf](https://github.com/crossterm-rs/crossterm/blob/05df63ab2bfcb6d91a0637dad196749ec4ff12f7/examples/perf.rs)

It takes 2m44s to finish the operation. Last line of the output:

```
Count: 10836505 None count: 10962315 Elapsed: 164.215568412s Too long processing: 0
```

* count = number of `Events` (not bytes, but events like `Event::Key(KeyCode::Char('a'))`
* none count = number of `Ok(false)` returned by `poll(0)`
* too long processing = number of `poll(0)` calls that took more than 5ms

### crossterm and socat

* [socat -s -b 2048 -u OPEN:../test.txt,ignoreeof EXEC:'cargo run --release --example perf',pty,setsid,ctty](https://github.com/crossterm-rs/crossterm/blob/05df63ab2bfcb6d91a0637dad196749ec4ff12f7/examples/perf.rs)

It takes 5.9s. Last line of the output:

```
Count: 10836505 None count: 0 Elapsed: 5.918270894s Too long processing: 0
```

Same explanation as in the previous example. The interesting part here is the `None count: 0`. `poll(0)` didn't return `Ok(false)` at all.

## crossterm's `poll` & `read`

* `Event` is parsed ANSI sequence (can be constructed from multiple bytes)
  * Example `Event::Key`, `Event::Mouse`, ...
* `read`
  * Reads an `Event` (not bytes) and blocks till an `Event` is available
* `poll`
  * Checks if an `Event` (not bytes) is available to avoid blocking with `read`
  * `mio::Poll` checks for available bytes to read, but not if it's an `Event` or not
  * `poll` internally ...
    * Checks if input is available via `mio::Poll`
    * Then these bytes are `read` & parsed
    * All parsed `Event`s are queued and `Ok(true)` is returned
    * If there's no `Event` and timeout didn't expire, it jumps to the beginning and does the same stuff with the `mio::Poll` & smaller timeout (leftover)
    * If the timeout expires, it returns `Ok(false)`

Simplified explanation how the upcoming version of the `crossterm` crate works.

## Profiling

I profiled/sampled the code and `poll` with zero timeout (= check & return immediately) returns `Ok(false)` many times. The number of `Ok(false)` occurences roughly equals to number of processed/parsed `Event`s. See the output from above example:

```
Count: 10836505 None count: 10962315 Elapsed: 164.215568412s Too long processing: 0
```

`10.8M` vs `10.9M`. I checked what's going on and the `mio::Poll` with zero timeouts says that there's nothing to read from the tty. So, there's probably nothing we can do right now. I can imagine that the terminal/tty can't catch up with the paste speed. I tried to increase the speed, etc. but it didn't help (negligible difference).

I also tried to modify the code to return immediately when we parse first `Event` (do not process the whole buffer), but the whole processing is slower. Even if it will be faster, it has nothing to do with `mio::Poll` returning _nothing to read_.

Check the attached screenshot:

Screenshot 2019-11-26 at 09 58 27

It's just `8.91s` sample, but enough to confirm that `mio::Poll` returns _nothing to read_ and when the time is being spent.

Weight vs Self Weight explanation to understand the meaning. Let's say you have a function `X` and it calls functions `Y`, `Z`. Weight represents time spent for `X` + `Y` + `Z` and Self Weight represents just time spent in `X`, but not in `Y` & `Z`.

## `poll` with different timeouts

For just first 1M events.

| Timeout | Count<sup>1</sup> | None count | Ratio <sup>2</sup> | Slow processing |
|:-- | --: | --: | --:| --: |
| `0.5ms` | `1_000_000` | `23_574` | 2.3574% | `0` |
| `1ms` | `1_000_000` | `13_416` | 1.3416% | `0` |
| `2ms` |  `1_000_000` | `6_352` | 0.6352% | `0` |
| `3ms` |  `1_000_000` | `4_648` | 0.4648% | `0` |
| `4ms` |  `1_000_000` | `2_983` | 0.2983% | `0` |
| `5ms` |  `1_000_000` | `2_649` | 0.2649% | <sup>3</sup> `1_175` |

<sup>1</sup> No 10MB as in other examples, just 1M of `Events`.
<sup>2</sup> none count / count
<sup>3</sup> That's fine as slow processing is defined as `> 5ms` in the `perf` example and the timeout is `5ms`.

## Platform

All above numbers come from the macOS. When I tried `poll(0)` on Linux, it's better. Let's check the whole 10MB test file and copy & paste method.

| Platform | Count | None count | Ratio | Total time |
| -- | --: | --: | --: | --: |
| macOS | `10_836_505` | `10_962_315` | 101% | 164.2s
| Linux | `10_836_505` | `2_884_389` | 27% | 6.4s |

Linux is way faster and `poll(0)` produces way less `Ok(false)` values.

## Recommendations

> `None` should only be returned if no event is available for polling; if "something unknown" is available, you can use `Event::Refresh` or `Event::Unknown`.

Frankly speaking, this is not possible. As you can see, `mio::Poll` says - there's nothing to read many many times. Which means that we don't know if we have "something unknown" or "nothing" - `mio::Poll` says "nothing".

One would say that we can use some very very small timeout, like 500 microseconds, but do we really want it? Also it's platform specific, so, this timeout in the `crossterm` backend should be `#[cfg(target_os = "linux")]`/`#[cfg(target_os = "macos")]`/... What is the acceptable ratio? < 1% which means 2ms timeout on macOS and 1ms on Linux which means 0.0076%? Another problem is that I can measure this on my computer, provide some numbers, but then it can behave differently (and surely will) on another computer.

I don't know `cursive` use cases, how fast it should be, what it can cause if we add a timeout, if we return `Event::Refresh` for every single `Ok(false)`, ...

When I look at the `Backend` trait ...

```rust
    /// Polls the backend for any input.
    ///
    /// Should return immediately.
    fn poll_event(&mut self) -> Option<Event>;
```

... there's no way to specify timeout. We will violate _Should return immediately._ if we will add a timeout. If we wont add it, there will be lot of `None` results. And again, the problem is in the `mio::Poll` which does return "nothing to read" quite often and it's platform specific.

@gyscos you should know what's going on now and where the problem is. What do you suggest?

zrzka:
One more thing. We do use `PollOpt::level` internally. Going to try `PollOpt::edge` (needs some rewrites) if it will be better or not.

zrzka:
Pragmatic approach & fix [TimonPost/crossterm_update..zrzka/crossterm-0.14-performance](https://github.com/TimonPost/cursive/compare/crossterm_update...zrzka:zrzka/crossterm-0.14-performance?expand=1).

src/backend/crossterm.rs
```diff
@@ -36,149 +155,96 @@ impl Backend {
...
+                Ok(event) => Some(self.map_key(event)),
+                Err(_) => panic!(),
+            },
+            _ => None,
+        }
     }
 
     fn finish(&mut self) {
-        input().disable_mouse_mode().unwrap();
-        cursor().show().unwrap();
```
vmedea: please don't forget to re-show the cursor !
(tried on Ubuntu)

src/backend/crossterm.rs
```diff
@@ -4,28 +4,147 @@
...
+            CKeyEvent {
+                modifiers: KeyModifiers::SHIFT,
+                code,
+            } => Event::Shift(Key::from(code)),
+
+            CKeyEvent {
+                modifiers: _,
+                code: KeyCode::Char(c),
+            } => Event::Char(c),
+            CKeyEvent { modifiers: _, code } => Event::Key(Key::from(code)),
```
vmedea: maybe
```rust
            CKeyEvent {
                modifiers: _,
                code: KeyCode::BackTab,
            } => Event::Shift(Key::Tab),
```
(this makes shift-tab work the same as with the ncurses backend, only tested on ubuntu)

TimonPost:
> One would say that we can use some very very small timeout, like 500 microseconds, but do we really want it?

So this boils down to - correct me if I am wrong @zrzka - do we wait a fraction of a millisecond, or do we spawn a thread which performs a block read and sent events over crossbeam channel back to the backend. This is for @gyscos to decide.

src/backend/crossterm.rs
TimonPost: Doesn't this already work? When I press SHIFT + TAB, crossterm sends a Backtap key, this BACK_TAP key is translated to TAB [over here](https://github.com/TimonPost/cursive/blob/crossterm_update/src/backend/crossterm.rs#L75). And will be [parsed here.](https://github.com/TimonPost/cursive/blob/crossterm_update/src/backend/crossterm.rs#L139)

src/backend/crossterm.rs
TimonPost: Didn't it work on ubuntu? It is supposed to so with this [statement](https://github.com/TimonPost/cursive/blob/crossterm_update/src/backend/crossterm.rs#L242).

zrzka:
Yes, it boils down to these two options. I don't like poll with timeout as it violates the docs and it's fragile - we have no idea what number should be set -> thread like termion

Third option, I'm going to look at mio, prepare minimal example and will consult it with mio developers. Didn't have time for this today.

src/backend/crossterm.rs
vmedea: it didn't work for me, shift-tab was interpreted as forward tab (both in the `keycode` example and in `radio`)

src/backend/crossterm.rs
vmedea: exactly; with current master, with crossterm backend, the cursor is visible after exiting each of the cursive demos, with this branch, it isn't
i've tested in gnome-terminal and inside tmux, same effect!

src/backend/crossterm.rs
```diff
@@ -302,10 +375,6 @@ where
             f(&Color::AnsiValue(16 + 36 * r + 6 * g + b))
         }
 
-        theme::Color::TerminalDefault => {
-            unimplemented!(
-                "I have to take a look at how reset has to work out"
-            );
-        }
+        theme::Color::TerminalDefault => f(&Color::Reset),
```
vmedea: it's nice that this works now

gyscos:
Thanks again for the thorough debugging!

So the main problem is that `mio` often returns that nothing can be read, even though there _is_ data in the tty buffer?

A blocking read in a separate thread is what termion does, but it's not _super_ clean (it may consumes one extra event after stopping the backend, so running multiple cursive loops consecutively is not always working as intended).

I'm not entirely against waiting ~1 or 2 ms: it's not _strictly_ returning instantly, but it's hopefully fast enough that it won't be noticeable/discernible from some heavy computation. It's especially tempting as a temporary solution until `mio` is fixed.

zrzka:
> So the main problem is that mio often returns that nothing can be read, even though there is data in the tty buffer?

Yep, but the _is data in the tty buffer_ here should be in quotes. To be sure, I'm going to prepare minimal example without cursive/crossterm to verify my findings and to have something for mio developers.

zrzka:
Minimal reproducible example without `cursive` & `crossterm` boilerplate in the `mio` repository here: https://github.com/tokio-rs/mio/issues/1171.

* Tried `mio` 0.6 & upcoming 0.7 - no difference
* With this issue, I'm not saying that it's the `mio` crate problem, it can be "this is how the stdin works" (slow, ...)

src/backend/crossterm.rs
TimonPost: I fixed the issue

src/backend/crossterm.rs
TimonPost: Fixed

src/backend/crossterm.rs
vmedea: yes this works now thank youu

TimonPost:
I am finishing up the last PR and examples in order to do the release. Shortly after the release I will update this PR, and remove the WIP-status.

TimonPost:
This PR is ready to me merges/reviewd. 0.14 is released.

TimonPost:
CC: @Shnatsel, this PR is not yet merged, we might discuss it here.

TimonPost:
Why should we use crossterm?

* Cursive does not have to depend on `term_size` to retrieve terminal size.
* Cursive can use resize events from crossterm instead of the signal polling thread. 
* No need for ncusus c dependencies. 
* Compilable and runnable both windows an Linux.

Earlier Concerns
* Performance
* Unstable

Revisiting Concerns
* Lots of improvements are done during the last few months. I would say performance is equal to termion (subjective argument). 
* Crossterm has been stabilized with 0.12,0.13,0.14, I think that it soon can be a 1.0.
* I merge and release issues bug fixes quickly.

Shnatsel:
I've opened an issue to discuss the choice of default backend further, see #411

On an unrelated note, does this fix #352?

TimonPost:
@gyscos  Can this PR be merged? All discussed things are handled.

vmedea:
> Why should we use crossterm as default

for true color support ! :rainbow: 

anyhow i've been using this for a while (on linux) and have found no issues, it works as expected

rustfmt.toml
```diff
@@ -1,5 +1,4 @@
 unstable_features = true
 max_width = 79
 reorder_imports = true
-fn_args_layout = "Tall"
```
vmedea: this appears to be a no-op change

TimonPost: Mmm, will give it a check.

gyscos:
Thanks! Will give it a try soon, should be able to get it merged before the new year.

rustfmt.toml
gyscos: The difference is the missing newline in the new file, let's just revert this.

src/backend/crossterm.rs
```diff
@@ -4,28 +4,188 @@
...
+            theme::Color::Dark(theme::BaseColor::Red) => Color::DarkRed,
+            theme::Color::Dark(theme::BaseColor::Green) => Color::DarkGreen,
+            theme::Color::Dark(theme::BaseColor::Yellow) => Color::DarkYellow,
+            theme::Color::Dark(theme::BaseColor::Blue) => Color::DarkBlue,
+            theme::Color::Dark(theme::BaseColor::Magenta) => {
+                Color::DarkMagenta
+            }
+            theme::Color::Dark(theme::BaseColor::Cyan) => Color::DarkCyan,
+            theme::Color::Dark(theme::BaseColor::White) => Color::Grey,
+            theme::Color::Light(theme::BaseColor::Black) => Color::Grey,
```
gyscos: Note that "dark white" and "light dark" are two different colors, index 7 and 8 respectively in the palette.

gyscos:
The current implementation writes to stdout (and I assume reads from stdin?).
Is it possible to work with `/dev/tty` instead? It will allow interactive applications that still use stdin/stdout, like the `fzf` tool, or the `vpv` example` (among others)

TimonPost:
Crissterm, on Linux, polls input from the tty https://github.com/crossterm-rs/crossterm/blob/master/src/event/sys/unix/file_descriptor.rs#L74

/dev/tty can't be used for windows systems, so I can't make it default in this implementation. That's why I used stdout. Can we somehow use an generic as buffer in cursive? That way the user can construct and create their own type of buffer

gyscos:
If we just want a difference between windows and Linux, we can do it with a `#cfg[]` in the backend code here?

TimonPost:
Sure, I can take a look. We can make unix systems use /dev/tty.

src/backend/crossterm.rs
vmedea: ye, this would be `Color::DarkGrey`

gyscos: Or maybe `Color::DarkWhite`?

TimonPost: Dark white does not exist in crossterm, this would just be grey. This should be DarkGrey as @vmedea  said.

gyscos: Ah, I was looking at the documentation, but I guess it's outdated:

https://docs.rs/crossterm/0.14.1/crossterm/style/enum.Color.html#platform-specific-notes

TimonPost: Thanks, I will have to update that.

gyscos:
There's still a few issues when redirecting `stdout` on linux (I think things like enabling the mouse are still sent to `stdout` instead of `/dev/tty`), but I'll merge it as-is and we'll fix it in another PR.

PR #408, opened 2019-12-01
Change tertiary color to white instead of grey

vmedea:
Change the tertiary color in the default theme to white instead of light grey.

Currently, the tertiary color is the same as the view background color, making it invisible as foreground color. This makes the Outset effect not work as (I think) intended.

Before:
a

After:
b


gyscos:
Whoops indeed! Thanks for the PR!

Issue #371, opened 2019-08-27
is multiline SelectView labels supported? 


sprajagopal:
First: is this a bug report? A suggestion? Or asking for help?
Asking for help 
# Problem description
How to create a selectview with multiline labels?

# Environment

* Operating system: Ubuntu 18.04
* Backend used: ncurses 
  * ncursew 6.1
* Current locale (run `locale` in a terminal).
```
LANG=en_IN
LANGUAGE=
LC_CTYPE="en_IN"
LC_NUMERIC="en_IN"
LC_TIME="en_IN"
LC_COLLATE="en_IN"
LC_MONETARY="en_IN"
LC_MESSAGES="en_IN"
LC_PAPER="en_IN"
LC_NAME="en_IN"
LC_ADDRESS="en_IN"
LC_TELEPHONE="en_IN"
LC_MEASUREMENT="en_IN"
LC_IDENTIFICATION="en_IN"
LC_ALL=

```
* Cursive version (from crates.io, from git, ...)
Cursive 0.12.0

sprajagopal:
@gyscos Awesome! Any temporary workaround you'd suggest for now?

PR #413, opened 2019-12-19
WIP: list_view with variable height views

Vrixyz:
Related to https://github.com/gyscos/cursive/issues/371

🚧 this pull request is not ready to be merged.

TODO:
- [ ] Have a way to know `ListView` items' height from its `draw` method: `ListChild` has no "size" property, we might want to mimic the behaviour from `LinearLayout` and its `Child` property
- [ ] Real tests
- [ ] Update documentation

### DIFF
--- a/examples/list_view_variable_height.rs
+++ b/examples/list_view_variable_height.rs
@@ -1,6 +1,7 @@
 use cursive::traits::*;
 use cursive::views::{
-    Checkbox, Dialog, EditView, LinearLayout, ListView, SelectView, TextArea, TextView,
+    Checkbox, Dialog, EditView, LinearLayout, ListView, SelectView, TextArea,
+    TextView,
 };
 use cursive::Cursive;
 
--- a/examples/list_view_variable_height.rs
+++ b/examples/list_view_variable_height.rs
@@ -1,0 +1,73 @@
+use cursive::align::HAlign;
+use cursive::traits::*;
+use cursive::views::{
+    Checkbox, Dialog, EditView, LinearLayout, ListView, SelectView, TextArea, TextView,
+};
+use cursive::Cursive;
+
+// This example uses a ListView.
+//
+// ListView can be used to build forms, with a list of inputs.
+
+fn main() {
+    let mut siv = Cursive::default();
+
+    siv.add_layer(
+        Dialog::new()
+            .title("Please fill out this form")
+            .button("Ok", |s| s.quit())
+            .content(
+                ListView::new()
+                    // Each child is a single-line view with a label
+                    .child("Name", TextArea::new().with_id("text"))
+                    .child(
+                        "Receive spam?",
+                        Checkbox::new().on_change(|s, checked| {
+                            // Enable/Disable the next field depending on this checkbox
+                            for name in &["email1", "email2"] {
+                                s.call_on_id(name, |view: &mut EditView| {
+                                    view.set_enabled(checked)
+                                });
+                                if checked {
+                                    s.focus_id("email1").unwrap();
+                                }
+                            }
+                        }),
+                    )
+                    .child(
+                        "Email",
+                        // Each child must have a height of 1 line,
+                        // but we can still combine multiple views!
+                        LinearLayout::horizontal()
+                            .child(
+                                EditView::new()
+                                    .disabled()
+                                    .with_id("email1")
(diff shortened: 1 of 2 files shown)


src/views/list_view.rs
```diff
@@ -281,60 +325,45 @@ impl View for ListView {
...
         for (i, child) in self.children.iter().enumerate() {
             match child {
                 ListChild::Row(ref label, ref view) => {
-                    printer.print((0, i), label);
+                    printer.print((0, yOffset), label);
                     view.draw(
-                        &printer.offset((offset, i)).focused(i == self.focus),
+                        &printer.offset((offset, yOffset)).focused(i == self.focus),
                     );
+                    yOffset += view.important_area(child.size).height();
```
Vrixyz: This is the line posing problem, important_area is not the correct function to call to get the proper height.

src/views/list_view.rs
```diff
@@ -222,6 +229,43 @@ impl ListView {
...
+            .iter_mut()
+            .filter_map(ListChild::view)
+            .map(|v| v.required_size(req))
+            .scan(&mut views_size, |s, size| {
+                let width = size.x;
+                if width > (**s).max_width {
+                    (**s).max_width = width;
+                }
+                (**s).sum_height += size.y;
+                println!("view_size: {:#?}", **s);
```
Vrixyz: To remove before merge

src/views/list_view.rs
```diff
@@ -222,6 +229,43 @@ impl ListView {
...
+            .children
+            .iter()
+            .map(ListChild::label)
+            .scan(&mut labels_size, |s, label| {
+                let width = label.width();
+                if width > (**s).max_width {
+                    (**s).max_width = width;
+                }
+                (**s).sum_height += 1;
+                //println!("{:#?}", **s);
```
Vrixyz: To remove before merge

examples/list_view_variable_height.rs
```diff
@@ -0,0 +1,73 @@
+use cursive::align::HAlign;
+use cursive::traits::*;
+use cursive::views::{
+    Checkbox, Dialog, EditView, LinearLayout, ListView, SelectView, TextArea, TextView,
+};
+use cursive::Cursive;
+
+// This example uses a ListView.
+//
+// ListView can be used to build forms, with a list of inputs.
```
Vrixyz: doc should be updated, and example could be simpler ?

src/views/list_view.rs
```diff
@@ -222,6 +229,43 @@ impl ListView {
...
+            .scan(&mut views_size, |s, size| {
+                let width = size.x;
+                if width > (**s).max_width {
+                    (**s).max_width = width;
+                }
+                (**s).sum_height += size.y;
+                println!("view_size: {:#?}", **s);
+                Some((**s).clone())
+            });
+        while scan.next().is_some() {}
```
Vrixyz: Not sure if that's really the **rust way™️**

gyscos: Could be `for _ in scan {}`?
But I think rather than the `.scan()`, a `for`-loop may be clearer.

```rust
        let mut labels_size = ColumSize { max_width: 0, max_height: 0 };
        for label in self.children.iter().map(ListChild::label) {
                let width = label.width();
                if width > labels_size.max_width {
                    labels_size.max_width = width;
                }
                labels_size.sum_height += 1;
        }
```

Or even simpler:

```rust
        // Convenient function to take the max of values on one axis, and the sum on the other axis
        let max_size = cursive::direction::Orientation::Vertical.stack(
            self.children.iter().map(|c| Vec2::new(c.label().width(), 1))
        );

        // Then it's easy to convert to a `ColumnSize` - but why even need a `ColumnSize` type vs `Vec2`?
        let labels_size = ColumnSize { max_width: max_size.x, max_height: max_size.y };
```

gyscos:
You'll also want to update the code that handles mouse events, to correctly compute the selected view:
https://github.com/gyscos/cursive/blob/6ce15f4c4ccb75fa721b31a3d1d6937198b24784/src/views/list_view.rs#L291

gyscos:
Sorry, closed by mistake.

gyscos:
Latest main branch should have working ListView with multi-line children.
Thanks again for the work, and sorry I didn't end up using this code.

PR #417, opened 2020-01-02
Fix scroll event conversion in crossterm backend

nytopop:
As it stands, `ScrollUp` events are interpreted as `WheelDown`; this pr fixes that.


gyscos:
Good catch! Thanks for the fix!

PR #422, opened 2020-01-27
Add Display & Eq trait for ObservedScreen

jwuensche:
Hi Alexandre :D 
this pull request will add a `Display` implementation for `ObservedScreen`s. 
##### Why
Recently I tried to perform snapshot tests of screens with the cursive puppet backend, but wasn't able to compare instances of `ObservedScreen`s to each other.
At the moment observed screens only provide to be outputted directly to stdout via `print_stdout`. This PR simply uses this implementation adjusted to work with a `Formatter` so we can implement the `Display` trait.
In combination with e.g. [insta](https://crates.io/crates/insta) this can easily be used to perform snapshot tests of cursive applications.
Also added `Eq` and `PartialEq` in case some tests require a screen to screen comparison without formatting the screens.

The old `print_stdout` is still available and is just calling `println!` using `Display`.

### DIFF
--- a/src/backend/puppet/observed.rs
+++ b/src/backend/puppet/observed.rs
@@ -84,12 +84,61 @@ impl ObservedCell {
 /// Puppet backend output
 ///
 /// Represents single frame.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Eq, PartialEq)]
 pub struct ObservedScreen {
     /// Size
     size: Vec2,
     /// Contents. Each cell can be set or empty.
     contents: Vec<Option<ObservedCell>>,
+}
+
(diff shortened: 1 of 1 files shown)


src/backend/puppet/observed.rs
```diff
@@ -7,6 +7,7 @@ use std::ops::Index;
 use std::ops::IndexMut;
 use std::rc::Rc;
 use std::string::ToString;
+use std::{fmt, fmt::Display, fmt::Formatter};
```
gyscos: Note: you can also do `use std::fmt::{self, Display, Formatter};`

gyscos:
Good idea - and thanks for the PR!

PR #423, opened 2020-01-31
Use mvaddnstr when printing repeated strs

mfontanini:
This change uses `mvaddnstr` rather than `mvaddstr` in n/pancurses' `print_at_rep` implementation. This is a slight optimization as `mvaddstr` will create a `CString` every time it's called, while `mvaddnstr` will only construct a single string and print it N times.

### DIFF
--- a/src/backend/curses/n.rs
+++ b/src/backend/curses/n.rs
@@ -380,12 +380,7 @@ impl backend::Backend for Backend {
 
     fn print_at_rep(&self, pos: Vec2, repetitions: usize, text: &str) {
         if repetitions > 0 {
-            ncurses::mvaddstr(pos.y as i32, pos.x as i32, text);
-            let mut dupes_left = repetitions - 1;
-            while dupes_left > 0 {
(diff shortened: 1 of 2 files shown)


gyscos:
Hi, and thanks for the PR!

I'm not sure this is what `mvaddnstr` does. It is a bit confusing, but [the documentation says](https://www.mkssoftware.com/docs/man3/curs_addstr.3.asp):

> The four routines with n as the last argument write at most n characters.

So I believe that the last argument is not a repetition count, but a maximum number of bytes to write out.

mfontanini:
lol that seems about right. Probably should have read the docs more before making the change. No wonder it was working faster!

gyscos:
That being said, I agree that it should be possible to generate a single `CString` and re-use it for all calls. Might need to dig deeper in `ncurses-rs` for that.

mfontanini:
Yeah, it's a bit overkill to allocate a string for every repetition but I'm not sure if you can get away with it.

PR #427, opened 2020-02-19
Fix Event::CtrlChar('z') binding for ncurses backend

danobi:
This patch fixes an off by one error where we don't handle ctrl-z
correctly.


gyscos:
Whoops, good catch! Thanks for the work!

Issue #428, opened 2020-02-26
Rendering an empty SelectView that uses popup() panics


compenguy:
# Problem description

This part works fine: non-popup `SelectView`s iterate over all the items, which in an empty `SelectView` is none, and has no issues.

The problem: popup `SelectView`s attempt to render the focused item, which in an empty `SelectView` causes a slicing panic trying to retrieve what isn't there.

# Environment

I reproduced this issue on Linux using the crossterm backend, but once you know what to look for, it should be pretty obvious by inspection (select_view.rs line 895 in the current revision of master).

# Solution

Expect a pull request soon for a fix that I've tested that uses `self.items.get(self.focus())` instead of `self.items[self.focus()]`, and only renders the label in the `Some()` case.

PR #429, opened 2020-02-26
Fix for cursive issue #428, "Rendering an empty SelectView..."

compenguy:
This change fixes cursive issue #428, caused by indexing into an empty
list.  The issue is corrected by using the .get() method to safely
handle the case where the requested index doesn't exist, and renders
no label in that case.

Fixes https://github.com/gyscos/cursive/issues/428


gyscos:
Wow - good catch! And thanks for the work!

PR #438, opened 2020-03-31
theme: Add PaletteColor::HighlightText

danobi:
You generally want to stick to default colors because of terminal color
schemes. If you don't, you risk getting suboptimal contrast and
difficult-to-read text. When you _do_ want to peg something to a color,
you need to fully specify the foreground and background to ensure you
have enough contrast for text to be readable.

Our project, below
(https://github.com/facebookincubator/resctl/tree/master/resctl/below),
wants to set a less jarring highlight color than red. In order to
effectively do this, we need a way to specify the highlighted text color
without causing a cascade of pegged colors. In other words, if we were
to manually set PaletteColor::View, we would need to manually set
everything PaletteColor::View is `ColorStyle`'d with, and so forth.

I realize this is an API breaking change. I spent some time considering
alternatives (such as detecting whether or not `HighlightText` is set,
and if not, use `View` as the highlighted text color). The problem with
the other options is that you'd need to:

  1) Document the fallback behavior
  2) Maintain backwards compat code paths

Seeing as cursive is pre-1.0 release, I figure this is the best way
forward.

### DIFF
--- a/src/theme/color_style.rs
+++ b/src/theme/color_style.rs
@@ -68,12 +68,12 @@ impl ColorStyle {
 
     /// Alternate text with highlight background.
     pub fn highlight() -> Self {
-        Self::new(PaletteColor::View, PaletteColor::Highlight)
+        Self::new(PaletteColor::HighlightText, PaletteColor::Highlight)
     }
 
     /// Highlight color for inactive views (not in focus).
     pub fn highlight_inactive() -> Self {
-        Self::new(PaletteColor::View, PaletteColor::HighlightInactive)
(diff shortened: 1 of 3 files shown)


gyscos:
Looks good! Thanks for the work!

gyscos:
Should it maybe also affect the foreground of `HighlightInactive`?

danobi:
> Should it maybe also affect the foreground of HighlightInactive?

Yeah probably. I'll update it

gyscos:
Thanks again!

Issue #445, opened 2020-04-24
Checkbox could use a better Interface, imo


sanbox-irl:
This is a suggestion / for comments. I'm comfortable writing up this PR myself (should be pretty simple).

The current Checkbox API lacks two features, and has a confusing API generally:
1. `check_bool(bool) -> Self` is missing. This means that if we create a box with an existing boolean variable, which the box should reflect, which we know when we create the box, we cannot make this using chainable calls, and instead must extract the checkbox into ugly calls like this:

```
        let mut amphibious_checkmark = Checkbox::new();
        if front.is_amphibious() {
            amphibious_checkmark.set_checked(self.front_is_amphibious);
        }
```

example of caller side chainable variable I propose:
```
     // within a larger chain!
Checkbox::new().check_bool(self.front_is_amphibious);
```

2. Equally, there is no equivalent for setting `disable`. To set disable, one must break their chain.

Is there any interest in PRs to add methods which do those things?

gyscos:
Thanks for the report!

Indeed, some chainable methods may be missing.

Note that often, the [`With`](https://docs.rs/cursive/0.14.0/cursive/trait.With.html) trait can be used to chain-ify any method:

```rust
use cursive::With;

let amphibious_checkmark = Checkbox::new().with_if(self.front_is_amphibious, Checkbox::check);
```

sanbox-irl:
Huh, I was completely unaware of the trait's existence! I'll put the PR on my calendar for this weekend :)! Have any thoughts on what the --- ah you just edited it. Perfect! Will do. Thanks for the uber quick response.

Unrelated question while I have you -- do you have any sort of showcase? I've been making some nice things with Cursive for work and I'd love to show off some screens :D

gyscos:
There's a smallish showcase in the Readme:
https://github.com/gyscos/cursive#showcases

Though it's fairly small and possibly outdated. I welcome any contribution!

PR #447, opened 2020-04-25
Expanded on API for Checkbox and for Enabled.

sanbox-irl:
Added a `with_x` variant to both, for easier chaining.

Because I added the `with_enabled` to the `impl_enabled` mod, this actually effect multiple widgets which can be disabled!

Addresses #445


gyscos:
Thanks for the woek!

sanbox-irl:
A mighty 50 seconds of work :P . Although it was nice to see how easy it is to add onto Cursive!

PR #448, opened 2020-04-25
Added impl_enabled to more widgets

sanbox-irl:
Added the impl_enabled macro to more widgets. This mostly just tidies up the codebase.

this PR has some of my own fork's history in it -- which is slightly messy. I'm not familiar enough with Git I think to make this cleaner. Sorry about that!


gyscos:
Looking good! thanks!

PR #449, opened 2020-04-26
Updated `Dialog` convencience methods: `test` and `info` to take `Into<StyledString>`

sanbox-irl:
This required no change to anything but Dialog, which previously took `Into<String>`, but TextboxView can take `Into<StyledString>` which will allow for slightly simpler phraseology. This allow for the following:

```
Dialog::text(StyledString::styled("Are you sure you want to do this?", Effect::Bold));
```
whereas before, the following was required:
```
Dialog::around(TextView::new(StyledString::styled("Are you sure you want to do this?", Effect::Bold)));
```

Because String implements `Into<StyledString>`, we lose no expressiveness, and this is not a breaking change.

(I think maybe this will make normal strings slower, but honestly, not sure about that -- I feel like optimizations will erase any differences here)

### DIFF
--- a/cursive-core/src/views/button.rs
+++ b/cursive-core/src/views/button.rs
@@ -5,7 +5,7 @@ use crate::rect::Rect;
 use crate::theme::ColorStyle;
 use crate::view::View;
 use crate::Vec2;
-use crate::{Cursive, Printer, With};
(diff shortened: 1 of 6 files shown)


gyscos:
Good idea! Thanks again!

PR #452, opened 2020-05-08
Fix the theme example

vmedea:
- Remove the inline feature detection, it doesn't work for features of dependencies. Instead use `required-features` in `Cargo.toml` for the  specific binary.

- The asset moved one level up relatively to `../../assets/style.toml`


gyscos:
Oh nice, thanks!

gyscos:
Re: the assets, I guess we could as well move it to the examples directory

PR #471, opened 2020-07-19
Fix text alignment when wrapped

### LINKED ISSUE #426: Wrapped text in TextView is not aligned properly
# Problem description

In a TextView, if a line of text is wrapped, then the line is not aligned properly:

```rust
use cursive::align::Align;
use cursive::view::Resizable;
use cursive::views::TextView;
use cursive::Cursive;

fn main() {
    let mut c = Cursive::default();
    c.add_layer(
        TextView::new("abcdef abcdefghij abcdef abcde")
            .align(Align::center())
            .fixed_size((16, 5)),
    );
    c.run();
}
```

Screenshot_20200216_182410

I expected the text to look like this:

Screenshot_20200216_182338

The same problem occurs with right-aligned text.

# Environment

* Operating system: Arch Linux
* Backend used: ncurses
  * If using ncurses on linux, ncurses version? 6.1
* Current locale: en_CA.UTF-8
* Cursive version: 0.14.0 from crates.io

j-tai:
This changes the `LinesIterator` to return a `width` that corresponds with the actual total length of the chunks, rather than returning the total available width, when a line is wrapped. This effectively reverts 20cb033b8d3c1edab642d9c068443cdd4d7c38b1 (I am not sure why that commit changed it).

Fixes #426

### DIFF
--- a/cursive-core/src/utils/lines/simple/lines_iterator.rs
+++ b/cursive-core/src/utils/lines/simple/lines_iterator.rs
@@ -67,8 +67,15 @@ impl<'a> Iterator for LinesIterator<'a> {
         let start = row.segments.first()?.start;
         let end = row.segments.last()?.end;
 
-        let width = row.width;
+        let spans::Row {
+            width, is_wrapped, ..
+        } = row;
 
-        Some(Row { start, end, width })
+        Some(Row {
+            start,
+            end,
+            width,
+            is_wrapped,
+        })
     }
 }
--- a/cursive-core/src/utils/lines/simple/row.rs
+++ b/cursive-core/src/utils/lines/simple/row.rs
@@ -10,8 +10,12 @@ pub struct Row {
     pub start: usize,
     /// End of the row (excluded)
     pub end: usize,
+
     /// Width of the row, in cells.
     pub width: usize,
+
+    /// Whether or not this text was wrapped onto the next line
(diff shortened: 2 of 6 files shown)


gyscos:
Hi, and thanks for the PR!

The problem https://github.com/gyscos/cursive/commit/20cb033b8d3c1edab642d9c068443cdd4d7c38b1 was trying to fix is to compute the size for `TextView` or `TextArea` when some likes have to be wrapped: we want the view to ask the full width in this case, even if the wrapped line is 1 cell shorter.
This is especially a problem given how these views handle their layout cache: if their final size is less than the requirements, then they assume they are not "constrained" and will not re-compute the layout if the available size increases. In practice this means that a `TextView` could get "stuck" while resizing the terminal, behaving as if the terminal was still small.
On a more minor details, it also means that a TextArea may have a surprising width depending on how the line wraps.

The fix at the time was to pretend wrapped lines were full-width, which fixed most problems, but is not the cleanest solution - and messes up the width computation for other purposes.

A better solution would be to include a flag when a row is wrapped, so that the view itself can decide to behave differently.

j-tai:
Does this look good?

gyscos:
Indeed! I just need to figure out how to update the `TextArea`, and it should be ready to go.

PR #485, opened 2020-08-10
update: restyle scroll bar

s97712:


gyscos:
Thanks for the PR!

A bit busy right now, should have time to check it out this weekend.

vmedea:
Would be nice to see a screenshot of the change before/after !

PR #493, opened 2020-08-21
menubar should bubble up events which are not handled

JoshMcguigan:
Hello, and thanks again for all your work on this crate. 

I found an issue where global callbacks were not being called while the menubar was focused. This PR fixes that by ensuring that events bubble up if they are not handled while the menu bar is focused.

I'm not sure this is the most direct way to accomplish this, since it seems if the menubar is focused we could somehow directly call the global events rather than relying on `View::on_event(&mut self.root, event.relativized((0, offset)))` to do it. But I wasn't sure exactly how to accomplish that. This PR does seem to work as-is.

### DIFF
--- a/cursive-core/src/cursive.rs
+++ b/cursive-core/src/cursive.rs
@@ -786,16 +786,25 @@ impl Cursive {
         }
 
         if self.menubar.receive_events() {
-            self.menubar.on_event(event).process(self);
-        } else {
-            let offset = if self.menubar.autohide { 0 } else { 1 };
-
-            let result =
-                View::on_event(&mut self.root, event.relativized((0, offset)));
+            let result = self.menubar.on_event(event.clone());
 
             if let EventResult::Consumed(Some(cb)) = result {
                 cb(self);
+                // If the menubar handles this event, we return
+                // early so the event isn't handled twice. Otherwise
+                // the event should bubble up, even if the menubar is
+                // active, to ensure global callbacks are called.
+                return;
             }
+        }
+
+        let offset = if self.menubar.autohide { 0 } else { 1 };
+
+        let result =
+            View::on_event(&mut self.root, event.relativized((0, offset)));
+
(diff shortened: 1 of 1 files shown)


gyscos:
Hi, and thanks for the PR!

I see the problem - since we moved global callbacks to the root view, it's hard to have an event trigger a global callback without sending it to the root view, where it could potentially be caught by the regular views - which would be weird when the menubar is focused (for example you could still write a message in an input box with a menu open).

The ideal long-term solution is probably to push the menubar into a regular view, and put _that_ behind the root-level `OnEventView`. But we're not there yet.

A workaround would be to somehow send the event to the root OnEventView, but _only_ for the callbacks, not for the view itself - but right now `OnEventView` doesn't really let you trigger the registered callbacks without poking the child view.

None of these two solutions are trivial, but the workaround is probably less up-front cost.

JoshMcguigan:
I think you are suggesting adding some mechanism to allow `OnEventView` to expose the registered callbacks? Perhaps something like `self.root.execute_callbacks_on(event)`? 

Would there be any restrictions on running the `BeforeChild` vs `AfterChild`? I think in practice for this use case we could get away with only running `AfterChild` since that is how the global callbacks are registered, but for simplicity I think it makes sense to run all callbacks.

gyscos:
> I think you are suggesting adding some mechanism to allow OnEventView to expose the registered callbacks? Perhaps something like self.root.execute_callbacks_on(event)?

Sorry for the delay - yes, this is exactly what I was thinking about.

> Would there be any restrictions on running the BeforeChild vs AfterChild? I think in practice for this use case we could get away with only running AfterChild since that is how the global callbacks are registered, but for simplicity I think it makes sense to run all callbacks.

Yeah we should run all callbacks here. Global callbacks default to AfterChild, but there's still `Cursive::set_on_pre_event` to register BeforeChild events.

PR #495, opened 2020-08-23
fix missing brace

poorie:
Fix mixing brace in documentation of tutorial 3


gyscos:
Whoops, thanks for noticing!

PR #496, opened 2020-08-25
BLT: Use bearlibterminal char function

mehmooda:
It already handles shift, capslock and user keymap
Leave fallback for when it doesnt work (CTRL pressed?)


gyscos:
Looks great, thanks for the PR!

Issue #433, opened 2020-03-18
Docs: explain how to use callbacks


njskalski:
Hi,

a huge part of Cursive I don't understand is Callbacks. What are they for? What is the advantage of returning a callback in response to consuming event?

Can you please explain it a little better in docs?

danobi:
Returning a callback is a bit strange but it's how you get access to the `Cursive` object while keeping the initial callback static.

gyscos:
Here's a brief summary of the idea behind the design behind `View::on_event` and `EventResult`:
* `View::on_event` takes a `&mut self` so it can modify the view reacting to the event.
* As reaction to the event, we want to potentially update the view tree; this would need a `&mut Cursive`
* But the root `Cursive` contains the view, so from `&mut Cursive` I could get another `&mut my_view`. This is impossible by rust standard. For example, what if I remove the view itself in the callback? The view cannot really be dropped since we're still in one of its methods.

The conclusion is that we cannot _easily_[1] take a `&mut Cursive` in `View::on_event`.
Instead, we can prepare "what we would do _if_ we had this `&mut Cursive`" - that's the `EventResult`. You split your event response in two parts: the first half (the body of `View::on_event`) that updates the current view, and the second part (the `EventResult`) that updates the `Cursive` root to modify the view tree. 

_Most of the times_ it's easy to split (one of the two halves may be simply empty). In more complex cases where you want to update the current view based on something in the view tree, you'll just need to somehow update the current view from the `EventResult` itself. This can be done by sharing a `Rc` with the callback, or by re-finding the current view from the root if it has an ID or something.

In the future I want to add a "ViewPath" to the `on_event` method, which describes the path in the view tree at the moment the event is dispatched. It would then be easy to give this path to the `EventResult` to find the current view again, without the need for a `NamedView`.

Another long-term goal I had in mind was `EventResult` not meant for the `Cursive` root, but for one of the parent views. This would potentially allow "sub-root" views to sandbox children, or enable more inter-views communications. This is still a vague idea and will obviously need a lot more design to get anywhere near usable.

I'll update a bit the documentation for `View::on_event` to make some of this clearer.

[1]: One convoluted way to have both would be to wrap every single view in some sort of `Rc + RefCell`; though it would need a deep refactor of how the view tree is visited for events. Also it might add some (probably small) overhead to every single view, including wrapper views, which is something I'm trying to avoid (wrapper views are a good idea as long as they are really cheap).

PR #501, opened 2020-09-20
Improved the `select` example.

Mckol:
## Introduction
This PR is an effect of my initial confusion with the intended way of handling `Callback`s, and aims to help prevent that from happening to others in future.

## Story time!
I wanted to add custom keybindings to a `SelectView`, so I looked through the examples and found the `select` example which looked like exactly what I needed. After replicating the `OnEventView` wrapper part of the code, the keybindings worked as far as navigating the `SelectView` went, but I noticed all the callbacks set with `SelectView.on_select()` were ignored with the custom keybindings but worked with the default arrow ones.  
I started reading into the documentation, and noticed the `SelectView.select_up()` method returned a `Callback`, and the documentation said it should be ran on the `Cursive`. I couldn't figure out how to get a reference to `Cursive` inside the closure in `OnEventView.on_event_inner()` to run it like the documentation example has shown (something along the lines of `callback(&mut siv)` ), so I tried using `OnEventView.on_event()`, then getting a referece to the `SelectView` by name, and using `Cursive.cb_sink()` to send the `Callback`s but apparently they weren't `Sync` so that didn't work out.

Not until having found #433 and [this comment in particular](https://github.com/gyscos/cursive/issues/433#issuecomment-606799956) did I realize that I should look at the documentation of `EventResult`, which, in turn, made me realize that I can just pass the `Callback` inside the `EventResult`.

## Summary
I thought it would be a good idea to create this small PR to hopefully help other Cursive beginners be less confused :)
It effectively doesn't change anything in this example as far as functionality goes, but it helps bring attention to this important way of handling `Callback`s.

### DIFF
--- a/examples/src/bin/select.rs
+++ b/examples/src/bin/select.rs
@@ -27,12 +27,12 @@ fn main() {
     // Let's override the `j` and `k` keys for navigation
     let select = OnEventView::new(select)
         .on_pre_event_inner('k', |s, _| {
-            s.select_up(1);
-            Some(EventResult::Consumed(None))
+            let cb = s.select_up(1);
+            Some(EventResult::Consumed(Some(cb)))
         })
         .on_pre_event_inner('j', |s, _| {
-            s.select_down(1);
-            Some(EventResult::Consumed(None))
+            let cb = s.select_down(1);
+            Some(EventResult::Consumed(Some(cb)))
         });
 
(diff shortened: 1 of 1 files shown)


gyscos:
Hi, and thanks for the PR!
Indeed, it's probably best practice to always forward the event result here, even if the example itself does not make use of it.

PR #502, opened 2020-09-20
new example: autocomplete search box

TianyiShi2001:
Thank you so much for making this crate that saved my life!

I'm making an app in which Google-like autocomplete search boxes are used often. While entering the query on the top of such a search box, filtered results should be generated below on the fly. 

It seems that there isn't a ready-to-use `View` for this, so I made one by assembling existing `Views`. I guess it might be helpful to add this to `examples`.

### DIFF
--- a/examples/src/bin/autocomplete.rs
+++ b/examples/src/bin/autocomplete.rs
@@ -37,10 +37,9 @@ fn main() {
                         // Center the text horizontally
                         .h_align(HAlign::Center)
                         .with_name("matches")
-                        .scrollable()
-                        .fixed_size((20, 10)),
+                        .scrollable(),
                 )
-                .fixed_width(25),
+                .fixed_height(10),
         )
         .button("Quit", Cursive::quit)
         .title("Where are you from?"),
--- a/examples/src/bin/autocomplete.rs
+++ b/examples/src/bin/autocomplete.rs
@@ -37,7 +37,9 @@ fn main() {
                         .on_submit(show_next_window)
                         // Center the text horizontally
                         .h_align(HAlign::Center)
-                        .with_name("matches"),
+                        .with_name("matches")
+                        .scrollable()
+                        .fixed_size((20, 10)),
                 )
                 .fixed_width(25),
         )
@@ -57,7 +59,7 @@ fn on_edit(siv: &mut Cursive, _content: &str, _cursor: usize) {
         let query = query.to_owned().to_lowercase();
         city.contains(&query)
     });
-    // Update the results with the filtered array of cities
+    // Update the `matches` view with the filtered array of cities
     siv.call_on_name("matches", |v: &mut SelectView| {
         v.clear();
         v.add_all_str(matches);
--- a/examples/src/bin/autocomplete.rs
+++ b/examples/src/bin/autocomplete.rs
@@ -22,7 +22,12 @@ fn main() {
         Dialog::around(
             LinearLayout::vertical()
                 // the input is on the top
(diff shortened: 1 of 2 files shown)


examples/src/bin/autocomplete.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// Update results according to the query
+fn on_edit(siv: &mut Cursive, _content: &str, _cursor: usize) {
+    // Get the query
+    let query = siv.find_name::<EditView>("query").unwrap().get_content();
+    // Filter cities with names that starts with the query string
+    let matches = CITIES
+        .lines()
+        .filter(|&city| query.chars().zip(city.chars()).all(|(c, d)| c == d));
```
gyscos, line 52: Maybe [`str::starts_with`](https://doc.rust-lang.org/std/string/struct.String.html#method.starts_with) could help here?

gyscos:
Looks great! Just a detail regarding the string prefix check, but it looks pretty good!

examples/src/bin/autocomplete.rs
gyscos: Also, maybe accepting more than just prefixes (using `contains()` instead) could help too?

TianyiShi2001: Thank you for the suggestion! I've changed it to `contains()`.

TianyiShi2001:
スクリーンショット 2020-09-22 8 29 08
Currently, when the `matches` view is focusing on my desired result, I need to press the down arrow key to move the 'global focus' from the query box to the `matches` view before I can submit the city. It would be nice if I can press `Enter` while I'm at the query box, and submit the focused item in the `matches view`, however, this won't work because the `.submit()` method is private:

```rust
fn on_submit(siv: &mut Cursive, _content: &str) {
    siv.call_on_name("matches", |v: &mut SelectView| {
        // wont' work; private associated function
        v.submit();
    });
}
```

Are there any workarounds? Or could more methods be made public?

TianyiShi2001:
`.selection()` might help. I'll try that

TianyiShi2001:
I've also made it possible to submit the city without moving the focus into the `matches` view. If matches are found, pressing "Enter" in the query box will submit the first match. If no matches are found, the value in the query box is submitted (not all people live in big cities)

TianyiShi2001:
I think it's ready to be merged :)

examples/src/bin/autocomplete.rs
```diff
@@ -0,0 +1,95 @@
...
+                    SelectView::new()
+                        // shows all cities by default
+                        .with_all_str(CITIES.lines())
+                        // Sets the callback for when "Enter" is pressed.
+                        .on_submit(show_next_window)
+                        // Center the text horizontally
+                        .h_align(HAlign::Center)
+                        .with_name("matches")
+                        .scrollable()
+                        .fixed_size((20, 10)),
```
gyscos, line 41: Do we actually care about the width? We could probably just `fixed_height(10)`?

gyscos:
Note re: the non-public status of `submit`: you can also call `.on_event()` with a return key to the view.

examples/src/bin/autocomplete.rs
TianyiShi2001: I borrowed these two lines from the other example, `select.rs`, which uses the same set of data. You are right, in this case, the width of the query may exceed 20, so it would be better not to fix the width.

gyscos:
Thanks again for all the work! :)

PR #503, opened 2020-09-21
add new example: stopwatch

TianyiShi2001:
A stopwatch is implemented in this example.

This example is not complete. I managed to make the textbox update the elapsed time every second, but haven't figured out how I can pause or stop the stopwatch. @gyscos any suggestions?


gyscos:
One solution is not to use a thread at all; instead, a structure will hold:
* The current `status` (paused or running)
* The current `elapsed` time
* The time of `last_update`.

Then:
* To read the time: if `status` is paused, return `elapsed`. Otherwise, return `elapsed + now - last_update`.
* To pause the timer, update `elapsed` with the formula above, then change the status.
* To resume the timer, just update the status.

TianyiShi2001:
Thank you for the help! I've now made the timer to start/pause/resume by responding to pressing "Space". It's the first time I implement a View so I ran into some problems.

I also want to implement `stop` and `on_stop` which mirror `SelectView`'s `submit` and `on_submit`. However, if I write

```rust
       fn stop(&mut self) -> EventResult {
            self.pause();
            let cb = self.on_stop.clone().unwrap();
            // We return a Callback Rc<|s| cb(s, &*v)>
            EventResult::Consumed(Some(Callback::from_fn(move |s| {
                cb(s, self.elapsed)
            })))
        }
```

I get the error: `` `self` has an anonymous lifetime `'_` but it needs to satisfy a `'static` lifetime requirement``. How can I get around this?

gyscos:
I think when you write `self.elapsed` in the `move` closure, it tries to move `self`.
Instead, define `let elapsed = self.elapsed;` before, since it's `Copy` it should let the closure move _just the elapsed time_, leaving `self` alone.

examples/src/bin/stopwatch.rs
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        /// Sets a callback to be used when `<Enter>` is pressed.
+        ///
+        /// Also happens if the user clicks an item.
+        ///
+        /// The elapsed time will be given to the callback.
+        ///
+        /// See also cursive::views::select_view::SelectView::on_submit
+        pub fn on_stop<F, R>(&mut self, cb: F)
```
gyscos, line 71: The usual convention for views would be to have `fn set_on_stop(&mut self, ...)`, and a chainable `fn on_stop(self, ...) -> Self`.

examples/src/bin/stopwatch.rs
```diff
@@ -0,0 +1,115 @@
...
+            printer.print((0, 0), &pretty(self.read()));
+        }
+
+        fn on_event(&mut self, event: Event) -> EventResult {
+            match event {
+                // pause/resume the timer when pressing "Space"
+                Event::Char(' ') => {
+                    self.pause_or_resume();
+                }
+                Event::Key(Key::Enter) if self.on_stop.is_some() => {
```
gyscos, line 100: Does that mean `Enter` does _not_ stop the timer if no `on_stop` callback was given?

TianyiShi2001:
> I think when you write `self.elapsed` in the `move` closure, it tries to move `self`.
> Instead, define `let elapsed = self.elapsed;` before, since it's `Copy` it should let the closure move _just the elapsed time_, leaving `self` alone.

Exactly! Thank you!

examples/src/bin/stopwatch.rs
TianyiShi2001: Thank you for spotting this! This has been fixed in https://github.com/gyscos/cursive/pull/503/commits/9cff3346d8d2baebfa874dafb94a5ad0ed56f594

examples/src/bin/stopwatch.rs
TianyiShi2001: Thanks! I've made both `set_on_stop` and `on_stop`.

TianyiShi2001:
Adding the 'lap time' functionality might make this example too complicated. Instead, I am going to implement a fully-fledged TUI clock utility in a separate crate.

examples/src/bin/stopwatch.rs
```diff
@@ -0,0 +1,198 @@
...
+        }
+
+        fn stop(&mut self) -> EventResult {
+            let stopwatch = &mut self.stopwatch;
+            if !stopwatch.paused {
+                stopwatch.pause();
+            }
+            let result = if self.on_stop.is_some() {
+                let cb = self.on_stop.clone().unwrap();
+                let stopwatch_data = self.stopwatch.clone(); // TODO: remove clone
```
TianyiShi2001, line 163: Almost done... Except in line 163 where I haven't figured out how not to clone

gyscos: Yeaah... the issue is that the callback will be called later, possibly after _removing this view_. So we can't depend on any data in the view itself; the callback must be self-supported.
The solution is to either clone the data as you are doing (making it standalone), or wrap it in a `Rc` shared by both the view and the callback.

TianyiShi2001: I just came up with another approach: use `mem::replace` to replace `self.stopwatch` with a new, zeroed `StopWatch`, while getting the ownership of the 'freshly' produced data, which can then passed to the closure.

TianyiShi2001:
I realized that `EventResult::with_cb` cannot act as the shortcut for EventResult::Consumed(Some(Callback::from_fn_once(...))) and I have to use the latter form. 

Now, this example pretty much works as I expected! Thank you so much for the guidance!

examples/src/bin/stopwatch.rs
gyscos: You can use [`std::mem::take`](https://doc.rust-lang.org/std/mem/fn.take.html) which does this automatically for "default" types.

gyscos:
Thanks!
A dedicated crate for more features sounds like a great idea. This way we can simplify the example to the minimum, and it'll be a teaser for your library :p

TianyiShi2001:
Thank you for all the help! I've simplified the example. Meanwhile, I've implemented the basic countdown timer in [my crate](https://github.com/TianyiShi2001/clock-cli-rs).

gyscos:
I think we can simplify a bit further the example - let's focus on one feature (a stopwatch) and look for the _minimal_ example demonstrating this.

For example: https://gist.github.com/gyscos/67910f22e304214eaef00157faecb2ce

TianyiShi2001:
Cool, I've adopted your changes. Thank you!

gyscos:
Great! I think you just need to rebase to resolve the readme conflict.

TianyiShi2001:
Looks good now

PR #508, opened 2020-10-06
Remove outdated comment from button.rs

robinkrahl:
Before commit f9c9e565189a32a30ff2b2d1fad594e07e78961b, this match
interpreted the key code 10 as the Enter key.  Since it now uses the
Key::Enter variant instead, the comment explaining the magic number is
no longer needed.


gyscos:
Whoops, good catch! Thanks for the PR!

PR #509, opened 2020-10-06
Implement Default for XY

robinkrahl:
This patch implements Default for XY<T> if T implements Default.  This
makes it easier to deal with types like XY<usize>.

### DIFF
--- a/cursive-core/src/xy.rs
+++ b/cursive-core/src/xy.rs
@@ -2,7 +2,7 @@ use crate::direction::Orientation;
 use std::iter;
 
 /// A generic structure with a value for each axis.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Hash)]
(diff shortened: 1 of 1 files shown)


gyscos:
Thanks for the PR!

I think in our case we can just add `Default` to the derived implementations of `XY` (along with `Debug`, `Clone`, ...).

robinkrahl:
You’re right!  I’ve updated the PR.  But why doesn’t the compiler complain about `T` not (necessarily) implementing `Default`?

gyscos:
These derives automatically carry the trait bound to any generic type. It's not always accurate (sometime you don't really depend on the generic type implementing the trait, see [a playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=950288f8f7b3590b116e2bd5c4ff3673)), but in many cases (including ours) it's actually what we want.

robinkrahl:
Wow, that really surprises me.  I’m sure I’ve written something like your example before, expecting that it would *always* implement Default.  Well, it’s always good to learn something new!

PR #510, opened 2020-10-06
Implement FromIterator for Style

robinkrahl:
This patch implements `FromIterator<&Style>` and `FromIterator<T: Into<Style>>` for `Style` to make it easier to programatically create styles.  Style already has a merge method, but it takes a slice instead of an iterator.


gyscos:
Looking good, thanks for the work!

PR #511, opened 2020-10-06
Mark XY::stack_{horizontal, vertical} as must_use

robinkrahl:
The method names `stack_horizontal` and `stack_vertical` don’t make it clear whether the methods modify self or return the modified version.  Therefore, it is easy to use them wrong if you don’t look at the documentation.  This patch adds the `must_use` attribute to both methods to make it easier to spot such mistakes.


gyscos:
Good idea!

PR #512, opened 2020-10-06
Implement FromIterator for SpannedString

robinkrahl:
This patch implements `FromIterator<SpannedString<T>>` for `SpannedString<T>` to make it easier to create strings programatically.  We could also use fold directly without extracting the first element, but that would require an additional allocation.

----

This should be the last PR for today.  Thanks for the reviews and merges!


gyscos:
Thanks again!

PR #520, opened 2020-10-28
Really minor fix

ChaboCode:
Now it looks a little bit prettier


gyscos:
Thanks for the PR! I think you need to update the link as well, a bit further in the file.

gyscos:
Thanks again for the work! :)

PR #524, opened 2020-11-26
Make SpannedString iterators double ended & exact sized

htejun:
So that users can, for example, do .spans().len() to determine the number of
spans.


gyscos:
Good idea - and thanks for the work!

PR #526, opened 2020-12-06
Add Color::None to be able to create styles with only foreground or background color

TumbleOwlee:
Hello,

I'm using Cursive in a personal project and stumbled upon a problem with the styled `SpannedString`. I'm using a `SelectView` to list piped content or the content of a file. For example, I piped the colored output of `pygmentize` into my application. To colorize the input, I'm parsing the color sequences and transform them to styled `SpannedString`s. Because in some cases the sequences specify only one color (e.g. only foreground color) but `cursive::theme::ColorStyle` requires both, I always set the missing color to `Color::TerminalDefault`. That's fine, but now the highlighting of the selection breaks.

A minimal example of the problem and the proposed solution:
```rust
use cursive::theme::*;
use cursive::utils::span::*;
use cursive::view::*;
use cursive::views::*;
use cursive::*;

fn main() {
    let mut siv = Cursive::default();
    siv.set_global_callback('q', |s| s.quit());
    siv.add_fullscreen_layer(
        SelectView::<String>::new()
            .with_name("select-view")
            .full_width()
            .full_height(),
    );

    let mut theme = siv.current_theme().clone();
    theme.palette.set_color("view", Color::TerminalDefault);
    theme.palette.set_color("primary", Color::Dark(BaseColor::White));
    theme.palette.set_color("highlight", Color::from_256colors(238));
    siv.set_theme(theme);

    let style_no_bg = Style {
        color: Some(ColorStyle::new(
            ColorType::Color(Color::Dark(BaseColor::Yellow)),
            ColorType::Color(Color::None),
        )),
        ..Default::default()
    };
    let style_default = Style {
        color: Some(ColorStyle::new(
            ColorType::Color(Color::Dark(BaseColor::Yellow)),
            ColorType::Color(Color::TerminalDefault),
        )),
        ..Default::default()
    };

    if let Some(mut v) = siv.find_name::<SelectView<String>>("select-view") {
        v.add_item_str("Plain text without any style");
        v.add_item(
            SpannedString::styled("Text with only fg color set", style_no_bg),
            "".into(),
        );
        v.add_item(
            SpannedString::styled("Text with bg color because you have to set one", style_default),
            "".into(),
        );
    }

    siv.run();
}
```

[asciicast](https://asciinema.org/a/377566)

As you can see, because I have to set a background color for the style, the background color set by the highlighting of the `SelectView` is overridden. To circumvent this, I added the `None` option. If this option is set for any color, the currently active color will be used instead. I also added support in all available backends.

I really hope I didn't miss anything. Many thanks in advance.

### DIFF
--- a/cursive-core/src/theme/color.rs
+++ b/cursive-core/src/theme/color.rs
@@ -97,7 +97,9 @@ pub enum Color {
     /// These 216 possible colors are part of the default color palette (256 colors).
     RgbLowRes(u8, u8, u8),
 
-    /// Unset
+    /// No color
+    ///
+    /// The color will be determined by the active color of the view.
     None,
 }
 
--- a/cursive-core/src/theme/color.rs
+++ b/cursive-core/src/theme/color.rs
@@ -97,10 +97,10 @@ pub enum Color {
     /// These 216 possible colors are part of the default color palette (256 colors).
     RgbLowRes(u8, u8, u8),
 
-    /// No color
+    /// Inherit the color of the parent 
     ///
-    /// The color will be determined by the active color of the view.
-    None,
+    /// The color will be set by the parent.
+    InheritParent,
 }
 
 impl Color {
--- a/cursive-core/src/theme/color.rs
+++ b/cursive-core/src/theme/color.rs
@@ -96,6 +96,9 @@ pub enum Color {
     ///
     /// These 216 possible colors are part of the default color palette (256 colors).
     RgbLowRes(u8, u8, u8),
+
+    /// Unset
+    None,
 }
 
 impl Color {
--- a/cursive/src/backends/blt.rs
+++ b/cursive/src/backends/blt.rs
@@ -343,7 +343,7 @@ fn colour_to_blt_colour(
     curr_clr: Color,
 ) -> BltColor {
     let clr = match clr {
-        Color::None => curr_clr,
+        Color::InheritParent => curr_clr,
         _ => clr,
     };
     let (r, g, b) = match clr {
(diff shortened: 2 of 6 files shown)


gyscos:
Thanks for the work - looks great!

Will have a better look soon when I get some time. Some bikeshedding may include changing the name of the variant to something like `InheritParent`.

TumbleOwlee:
Thanks. Because my choice was really bad and "InheritParent" is far better to make clear what it actually does, I changed it for the moment. If I have done any mistakes, please feel free to let me know and I will improve on it.

gyscos:
Thinking about it more, I am wondering whether it makes more sense to do this at the `Color` level, or at the `ColorType` level.

I think moving it to `ColorType` would be more coherent: a `Color` would actually represent a single color, and `ColorType` would be a higher-level abstraction on how to obtain such a `Color`.

It also means maybe the backends aren't the ideal place to implement the behaviour itself; it would be done before that, maybe in the `Printer`. `ColorType::resolve` would be updated to take the current color as argument; `ColorStyle::resolve` would be updated to take the current color _pair_ as argument; and the `Printer` would maintain the state of current color pair to feed this.

gyscos:
I started an experiment in a branch if you want an example of what it could look like:
https://github.com/gyscos/cursive/commit/4d124ff893eaa6fa11ffe779548a5dcccb2a58a0

TumbleOwlee:
Hey, first of all, sorry, I hadn't much time the last couple of days. I checked your branch, the changes and applied it in my toy project. It works like a charm. I also checked every supported backend. Yours is by far the more elegant solution. Thanks! 
In this state, what are your plans? I believe this makes my request obsolete and can be closed. Will you merge your branch in the near future?

gyscos:
<MNhttps://github.argo.ai/argoai/labeling/pull/8792/files#diff-x

PR #535, opened 2021-01-13
select_view: Fix StyledString selection highlighting

### LINKED ISSUE #481: StyledString breaks SelectView highlighting
> First: is this a bug report? A suggestion? Or asking for help?

Bug

# Problem description

`StyledString`s causes a break in `SelectView` highlighting.

Here's a diff on one of the example programs:
```diff
iff --git a/examples/select.rs b/examples/select.rs
index 7d70021..4e4d57f 100644
--- a/examples/select.rs
+++ b/examples/select.rs
@@ -3,6 +3,8 @@ use cursive::event::EventResult;
 use cursive::traits::*;
 use cursive::views::{Dialog, OnEventView, SelectView, TextView};
 use cursive::Cursive;
+use cursive::utils::markup::StyledString;
+use cursive::theme::{Color, BaseColor};

 // We'll use a SelectView here.
 //
@@ -16,13 +18,10 @@ fn main() {
         // Use keyboard to jump to the pressed letters
         .autojump();

-    // Read the list of cities from separate file, and fill the view with it.
-    // (We include the file at compile-time to avoid runtime read errors.)
     let content = include_str!("../assets/cities.txt");
-    select.add_all_str(content.lines());
-
-    // Sets the callback for when "Enter" is pressed.
-    select.set_on_submit(show_next_window);
+    for line in content.lines() {
+        select.add_item(StyledString::styled(line, Color::Light(BaseColor::Green)), 0);
+    }

     // Let's override the `j` and `k` keys for navigation
     let select = OnEventView::new(select)
@@ -46,13 +45,3 @@ fn main() {

     siv.run();
 }
-
-// Let's put the callback in a separate function to keep it clean,
-// but it's not required.
-fn show_next_window(siv: &mut Cursive, city: &str) {
-    siv.pop_layer();
-    let text = format!("{} is a great city!", city);
-    siv.add_layer(
-        Dialog::around(TextView::new(text)).button("Quit", |s| s.quit()),
-    );
-}
```

Screenshot:
color highlight

# Environment

> * Operating system: linux distribution, windows, macOS, ...

fedora 32

> * Backend used: ncurses (the default one), pancurses, termion, ...
>  * If using ncurses on linux, ncurses version? (run `grep "define NCURSES_VERSION " /usr/include/curses.h`)

ncurses:
```
$ grep "define NCURSES_VERSION " /usr/include/curses.h
#define NCURSES_VERSION "6.1"
```

> * Current locale (run `locale` in a terminal).
```
$ locale
LANG=en_US.UTF-8
LC_CTYPE=en_US.UTF-8
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_PAPER="en_US.UTF-8"
LC_NAME="en_US.UTF-8"
LC_ADDRESS="en_US.UTF-8"
LC_TELEPHONE="en_US.UTF-8"
LC_MEASUREMENT="en_US.UTF-8"
LC_IDENTIFICATION="en_US.UTF-8"
LC_ALL=
```

> * Cursive version (from crates.io, from git, ...)

Built from master

danobi:
Before, highlighting would break if the focused item had custom styling.
THis was because the background color of the StyledString would not
necessarily match highlighting color.

This commit overwrites StyledString background colors to the
highlighting color if the item is focused.

This closes #481 .

### DIFF
--- a/cursive-core/src/views/select_view.rs
+++ b/cursive-core/src/views/select_view.rs
@@ -5,7 +5,7 @@ use crate::event::{
 };
 use crate::menu::MenuTree;
 use crate::rect::Rect;
-use crate::theme::ColorStyle;
+use crate::theme::{ColorStyle, PaletteColor};
 use crate::utils::markup::StyledString;
 use crate::view::{Position, View};
 use crate::views::MenuPopup;
@@ -446,7 +446,35 @@ impl<T: 'static> SelectView<T> {
         let l = self.items[i].label.width();
         let x = self.align.h.get_offset(l, printer.size.x);
         printer.print_hline((0, 0), x, " ");
-        printer.print_styled((x, 0), (&self.items[i].label).into());
+
+        // Overwrite the focused item's background color so that it matches the highlighting
+        // color. This prevents the focused text from having inconsistent highlighting color.
+        if i == self.focus() {
+            let styled_string = StyledString::with_spans(
+                self.items[i].label.source(),
+                self.items[i]
+                    .label
+                    .spans_raw()
+                    .iter()
+                    .map(|span| {
+                        let mut new_span = span.clone();
+                        if let Some(style) = span.attr.color {
+                            new_span.attr.color = Some(ColorStyle::new(
+                                style.front,
+                                PaletteColor::Highlight,
+                            ));
+                        }
+
+                        new_span
+                    })
+                    .collect(),
+            );
+
+            printer.print_styled((x, 0), (&styled_string).into());
+        } else {
+            printer.print_styled((x, 0), (&self.items[i].label).into());
+        }
+
         if l < printer.size.x {
             assert!((l + x) <= printer.size.x);
             printer.print_hline((x + l, 0), printer.size.x - (l + x), " ");


danobi:
With the following diff applied:
```diff
diff --git a/examples/src/bin/select.rs b/examples/src/bin/select.rs
index 4099dde..4b7cf77 100644
--- a/examples/src/bin/select.rs
+++ b/examples/src/bin/select.rs
@@ -1,6 +1,8 @@
 use cursive::align::HAlign;
 use cursive::event::EventResult;
+use cursive::theme::{BaseColor, Color};
 use cursive::traits::*;
+use cursive::utils::markup::StyledString;
 use cursive::views::{Dialog, OnEventView, SelectView, TextView};
 use cursive::Cursive;

@@ -16,13 +18,13 @@ fn main() {
         // Use keyboard to jump to the pressed letters
         .autojump();

-    // Read the list of cities from separate file, and fill the view with it.
-    // (We include the file at compile-time to avoid runtime read errors.)
     let content = include_str!("../../assets/cities.txt");
-    select.add_all_str(content.lines());
-
-    // Sets the callback for when "Enter" is pressed.
-    select.set_on_submit(show_next_window);
+    for line in content.lines() {
+        select.add_item(
+            StyledString::styled(line, Color::Light(BaseColor::Green)),
+            0,
+        );
+    }

     // Let's override the `j` and `k` keys for navigation
     let select = OnEventView::new(select)
@@ -46,13 +48,3 @@ fn main() {

     siv.run();
 }
-
-// Let's put the callback in a separate function to keep it clean,
-// but it's not required.
-fn show_next_window(siv: &mut Cursive, city: &str) {
-    siv.pop_layer();
-    let text = format!("{} is a great city!", city);
-    siv.add_layer(
-        Dialog::around(TextView::new(text)).button("Quit", |s| s.quit()),
-    );
-}
```

Before:
before

After:
after

lnyng:
I think the root cause of this issue is that when converting `Color` to `ColorStyle`, the current `PaletteColor::View` is used as the background: [code](https://github.com/gyscos/cursive/blob/main/cursive-core/src/theme/color_style.rs#L88-L92).

However, the intention is to have a transparent background, i.e. delay background color resolution until the moment of printing. In this particular case, the background color should have been `PaletteColor::Highlight` for selected items.

I think a real solution is to allow `ColorStyle` to have background and foreground colors as `Option` and only resolve at print time. But this diff as a workaround also works.

gyscos:
I finally merged the `inherit_parent_color_type` branch, which brings `ColorType::InheritParent`, which may help here.

gyscos:
I also just updated the `impl From<T> for ColorStyle` implementations, so it forwards the calls to `ColorStyle::front`, which uses `ColorType::InheritParent` as background color. The net result is that specifying a single color as a style now properly leaves the background "unset", and the patched `select` example from the first post now works as expected.

danobi:
@gyscos thanks! Works great

PR #541, opened 2021-01-31
Add getters & other utility methods to Dialog

kmeisthax:
This PR adds a handful of new methods to allow access to internal Dialog properties that are currently write-only. It also adds some setters for previously read-only properties. The whole list is as follows:

- `replace_content`, which allows one to take the contents of a `Dialog` without destroying the `Dialog`
- `get_h_align`, which allows reading the current button row `HAlign`
- `get_title`, which allows reading the current title
- `get_title_position`, which allows reading the current title `HAlign`
- `get_padding`, `get_padding_lrtb`, `get_padding_left`, `get_padding_top`, `get_padding_right`, and `get_padding_bottom`, which allow reading the padding in every way you can currently set it
- `iter_buttons` for immutably reading the buttons on a `Dialog`
- `iter_buttons_mut`, which is intended to replace the now-deprecated `buttons_mut` as usual Rust convention is to have `iter` in the name of methods that yield values as an iterator
- `set_focus`, which allows stealing `Dialog` focus. (We clamp the values given to us for safety.)

You're free to veto any of these changes if you think they're excessive. (I suspect renaming `buttons_mut` might be going too far.)

### DIFF
--- a/cursive-core/src/views/dialog.rs
+++ b/cursive-core/src/views/dialog.rs
@@ -11,7 +11,6 @@ use crate::{
 };
 use std::cell::Cell;
 use std::cmp::{max, min};
-use std::mem::replace;
 use unicode_width::UnicodeWidthStr;
 
 /// Identifies currently focused element in [`Dialog`].
@@ -167,9 +166,12 @@ impl Dialog {
     /// Previous content will be returned.
     pub fn set_content<V: IntoBoxedView>(&mut self, view: V) -> Box<dyn View> {
         self.invalidate();
-        replace(&mut self.content, LastSizeView::new(BoxedView::boxed(view)))
-            .view
-            .unwrap()
+        std::mem::replace(
+            &mut self.content,
+            LastSizeView::new(BoxedView::boxed(view)),
+        )
+        .view
+        .unwrap()
     }
 
     /// Convenient method to create a dialog with a simple text content.
--- a/cursive-core/src/views/dialog.rs
+++ b/cursive-core/src/views/dialog.rs
@@ -307,6 +307,11 @@ impl Dialog {
         self.invalidate();
     }
 
+    /// Get the title of the dialog.
+    pub fn get_title(&self) -> &str {
+        &self.title
+    }
+
     /// Sets the horizontal position of the title in the dialog.
     /// The default position is `HAlign::Center`
     pub fn title_position(self, align: HAlign) -> Self {
--- a/cursive-core/src/views/dialog.rs
+++ b/cursive-core/src/views/dialog.rs
@@ -357,18 +357,6 @@ impl Dialog {
         self.padding(Margins::lrtb(left, right, top, bottom))
     }
 
-    /// Gets the padding in the dialog.
-    ///
-    /// Returns Left, Right, Tob, Bottom fields as a tuple.
-    pub fn get_padding_lrtb(&self) -> (usize, usize, usize, usize) {
-        (
-            self.padding.left,
-            self.padding.right,
-            self.padding.top,
-            self.padding.bottom,
-        )
-    }
-
     /// Sets the padding in the dialog (around content and buttons).
     ///
     /// Chainable variant.
@@ -386,11 +374,6 @@ impl Dialog {
         self.padding.top = padding;
     }
 
-    /// Gets the top padding in the dialog (under the title).
-    pub fn get_padding_top(&mut self) -> usize {
-        self.padding.top
-    }
-
     /// Sets the bottom padding in the dialog (under buttons).
     pub fn padding_bottom(self, padding: usize) -> Self {
         self.with(|s| s.set_padding_bottom(padding))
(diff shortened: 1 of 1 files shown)


cursive-core/src/views/dialog.rs
```diff
@@ -167,6 +168,19 @@ impl Dialog {
...
 
+    /// Replace the content for this dialog.
+    ///
+    /// Previous content will be returned as a `BoxedView`.
+    pub fn replace_content<V: IntoBoxedView>(&mut self, view: V) -> BoxedView {
+        let old_content = replace(
+            &mut self.content,
+            LastSizeView::new(BoxedView::boxed(view)),
+        );
+        self.invalidate();
```
gyscos, line 179: We could first invalidate, then replace the content (so no need for a temp variable).
We also may want to call `BoxedView::unwrap` on the result and return a `Box<dyn View>` - this is what `StackView` does currently when poping layers.

We could also do that in the `set_content` method, we might not need a dedicated function.

cursive-core/src/views/dialog.rs
```diff
@@ -349,6 +395,11 @@ impl Dialog {
         self.padding.top = padding;
     }
 
+    /// Gets the top padding in the dialog (under the title).
+    pub fn get_padding_top(&mut self) -> usize {
```
gyscos, line 399: Not sure we need all these, could just let the user do `.get_padding().top`.

cursive-core/src/views/dialog.rs
```diff
@@ -379,17 +440,58 @@ impl Dialog {
...
+    /// attempting to focus a button that no longer exists will instead focus
+    /// one that does (or the content, if no buttons exist).
+    pub fn set_focus(&mut self, new_focus: DialogFocus) {
+        self.focus = match new_focus {
+            DialogFocus::Content => DialogFocus::Content,
+            DialogFocus::Button(_) if self.buttons.is_empty() => {
+                DialogFocus::Content
+            }
+            DialogFocus::Button(c) => {
+                DialogFocus::Button(max(c, self.buttons.len()))
```
gyscos, line 490: You probably want the `min` instead of `max` here.

cursive-core/src/views/dialog.rs
```diff
@@ -167,6 +168,19 @@ impl Dialog {
         self.invalidate();
     }
 
+    /// Replace the content for this dialog.
+    ///
+    /// Previous content will be returned as a `BoxedView`.
+    pub fn replace_content<V: IntoBoxedView>(&mut self, view: V) -> BoxedView {
+        let old_content = replace(
```
gyscos, line 175: I think we can use the full name `std::mem::replace` here without importing it first.

cursive-core/src/views/dialog.rs
```diff
@@ -379,17 +440,58 @@ impl Dialog {
         self.padding.right = padding;
     }
 
-    /// Returns an iterator on this buttons for this dialog.
-    pub fn buttons_mut(&mut self) -> impl Iterator<Item = &mut Button> {
```
gyscos, line 383: Between `buttons` / `buttons_mut` and `iter_buttons` / `iter_buttons_mut`, I think I do prefer `buttons`/`buttons_mut` - using simple plurals for iteration is the traditional approach. The only drawback is a potential conflict if we want a `buttons` chainable setters for an iterator of buttons, but I don't think it will happen.

I do agree we might want a non-mutable iterator version.

gyscos (commented):
Thanks for the PR! Just a few comments.

cursive-core/src/views/dialog.rs
```diff
@@ -332,6 +366,18 @@ impl Dialog {
         self.padding(Margins::lrtb(left, right, top, bottom))
     }
 
+    /// Gets the padding in the dialog.
+    ///
+    /// Returns Left, Right, Tob, Bottom fields as a tuple.
+    pub fn get_padding_lrtb(&self) -> (usize, usize, usize, usize) {
```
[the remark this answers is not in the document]
kmeisthax: I should point out that we already have `get_padding` to do that. I added an LRTB version for parity with the setters - if you don't think it's necessary, I'll just remove it.

cursive-core/src/views/dialog.rs
kmeisthax: I implemented this and got rid of `replace_content`, although I should point out that changing the return type is technically a breaking change.

cursive-core/src/views/dialog.rs
kmeisthax: I added these for parity with the setters; same thing with `get_padding_lrtb`. If you want I can remove these.

cursive-core/src/views/dialog.rs
gyscos: True, but I think it's worth the version bump to avoid having two similar functions.

gyscos (commented):
I don't think we particularly need parity between the setters and the getters.
In particular, a bunch of "convenient setters" exist because they simplify some common tasks, which doesn't need the same degree of simplification on the getter side.
* `padding_lrtb()` was mostly added to replace the previous ambiguous `From<(usize,usize,usize,usize)> for Margins` that has since been removed. It avoids having to explicitly name the `Margins` type to call `Margings::lrtb`. On the getter side however, there is no need to specify the full type name to retrieve a field or two.
* `padding_top` and the others allow setting a single padding field - the alternative would have been to first retrieve the full padding object, change only one field, then set it back. Here too, the getter side doesn't have this issue, as fetching a single field from the returned `Margins` struct is quite simple already.

gyscos:
Looks great, thanks again for all the work! :)

PR #546, opened 2021-02-06
Added label to checkbox

smallB007:
Mostly checkboxes are accompanied by labels. This feature seems to be missing from the current implementation.


gyscos:
Hi, and thanks for the work!
In general it should be easy to achieve the same using a horizontal LinearLayout with a checkbox and a label. Among other things, this naturally let users chose to put the label to the left or right of the checkbox, and supports styled text.

smallB007:
Yes, I agree with what you're saying.

PR #554, opened 2021-02-15
Added possibility to change based on flag during creation of a HideableView if it is hidden or not

smallB007:


gyscos:
Hi, and thanks for the work!

This indeed looks like a gap in the API. Good catch!
On a smaller detail, I think we should name this method `visible(self, visible: bool)` to go with `set_visible`. And it could simply call `self.with(|s| s.set_visible(visible))`, without the need to branch there.

gyscos:
Thanks again!

smallB007:
And thank you for your great work!

PR #577, opened 2021-05-05
Disable default enabled features

TimonPost:
When running crossterm-backend with this command, on linux, I will get an error that ncurses is not installed. By adding this, it works fine.


gyscos:
Indeed, thanks for the work!

PR #579, opened 2021-05-15
Only render visible content in TextView

### LINKED ISSUE #578: Application get's slow when displaying a large amount of text
When I display a large amount of text inside of an TextView, the Application gets slow. Is there any way to render only the portion of the text that get's displayed on the screen?

FliegendeWurst:
This should fix #578.


gyscos:
Thanks for the work!

PR #581, opened 2021-05-17
Use non-blocking I/O to fix #563

### LINKED ISSUE #563: [BUG] Termion backend drops some characters for subsequent reads from stdin
**Describe the bug**

When using the termion back-end and then reading from stdin, up to two characters are dropped.

**To Reproduce**

Simple modification of the pause example at https://github.com/gyscos/cursive/blob/main/examples/src/bin/pause.rs:

```rust
fn main() {
    let mut siv = cursive::termion();
    siv.add_layer(
        cursive::views::Dialog::text("Please write your message.")
            .button("Ok", |s| s.quit()),
    );
    siv.run();

    println!("Enter your message here:");
    let mut line = String::new();
    std::io::stdin().read_line(&mut line).unwrap();

    println!("Your message was: {:?}", line);
}
```

Expected behavior:

```
Enter your message here:
hello
Your message was: "hello\n"
```

Actual behavior:

```
Enter your message here:
hello
Your message was: "llo\n"
```

Note that the first two characters of the provided message have been dropped.

**Environment**
* Operating system used: macOS 10.15.7
* Backend used: `termion`
* Current locale:
```
LANG="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_CTYPE="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_ALL=
```
* Cursive version: 0.16.3

**Analysis**

I believe the main problem is this loop in combination with the implementation of termion:

https://github.com/gyscos/cursive/blob/7cf597f9bbe081fa3b37c5d5b8316c5298004a31/cursive/src/backends/termion.rs#L93-L108

When you ask termion for the next event, it reads up to two bytes, which is why specifically up to two characters are dropped:

https://github.com/redox-os/termion/blob/dce5e7500fd709987f9bf8f3911e4daa61d0ad14/src/input.rs#L60-L85

Even after the `Cursive` instance is dropped, the input thread is still running in the background. Once data has been written to stdin, this thread will then read up to two bytes and attempt to run `input_sender.send(event)`, which will fail and then cause the thread to terminate. The calling code may then attempt to read from stdin, but the bytes are already gone.

The ideal fix would be to terminate the input thread once the Cursive instance has been dropped, but I can't figure out a good way to do that, since the call to `events.next()` is fundamentally blocking. It might be possible to use a variant of `select` to poll the input to see if it has data before calling `events.next()`, but that seems complicated and non-portable.

I haven't been able to figure out a good workaround for user code. In particular, I expected that creating the Termion back-end with `/dev/tty` would fix the issue, but it doesn't seem to.

Termion has an item called `termion::async_stdin`, which might be relevant towards fixing this issue.

arxanas:
It turns out that Termion is already capable of handling I/O without consuming too much input. The hint is that it expects non-blocking input, as per this snippet:

```
        let mut buf = [0u8; 2];
        let res = match source.read(&mut buf) {
            Ok(0) => return None,
            Ok(1) => {
                match buf[0] {
                    b'\x1B' => Ok((Event::Key(Key::Esc), vec![b'\x1B'])),
                    c => parse_event(c, &mut source.bytes()),
                }
            }
            Ok(2) => {
```

One would expect a `read` call not to return `0` under normal circumstances. It could return `0` e.g. on <kbd>ctrl</kbd>+<kbd>d</kbd> (not really useful in an interactive context), or if a non-blocking read did not have any available data.

Tested using the program in https://github.com/gyscos/cursive/issues/563. Before this commit, it would drop the first two characters:

```
Enter your message here:
asfdasdf
Your message was: "fdasdf\n"
```

After this commit, it returns all the characters

```
Enter your message here:
asdfasdf  
Your message was: "asdfasdf\n"
```

### DIFF
--- a/cursive/src/backends/termion.rs
+++ b/cursive/src/backends/termion.rs
@@ -6,13 +6,13 @@
 
 pub use termion;
 
-use crossbeam_channel::{self, select, Receiver};
+use crossbeam_channel::{self, Receiver};
 use termion::color as tcolor;
 use termion::event::Event as TEvent;
 use termion::event::Key as TKey;
 use termion::event::MouseButton as TMouseButton;
 use termion::event::MouseEvent as TMouseEvent;
-use termion::input::{MouseTerminal, TermRead};
+use termion::input::{Events, MouseTerminal, TermRead};
 use termion::raw::{IntoRawMode, RawTerminal};
 use termion::screen::AlternateScreen;
 use termion::style as tstyle;
@@ -28,7 +28,6 @@ use std::fs::File;
 use std::io::{BufWriter, Write};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
-use std::thread;
 
 /// Backend using termion
 pub struct Backend {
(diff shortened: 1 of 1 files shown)


arxanas:
I suppose that if we want to commit to Unix-only (which seems reasonable, since Termion is Unix-only?), then we can omit the background event-reading thread altogether. We can just poll for events in a regular non-blocking fashion. I suspect the code leaks memory for the blocking input case, since we will poll for events much faster than we will produce them.

gyscos:
Hi, and thanks for the work! It's awesome to have actual non-blocking IO for the input, thanks for spending the time to find out how it was done! :)

I think restricting to Unix and relying on non-blocking-ness is totally fine, especially for termion which, as you said, doesn't support anything else anyway. Non-blocking input should map nicely to the current `Backend` trait, and a simpler implementation is always welcome.

I'm also still uneasy about the "read request channel" thing - last time it lead to a loads of issues, including poor performance, deadlocks, and overall making things more complicated than they should. The version in this PR looks better, but if we can avoid it, I say we should.

If/When we bring this to crossterm, we'll want to find a solution that applies to windows as well. I _hope_ we can find a way to make windows non-blocking, so we could remove the input thread there too.
It seems [`kbhit`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/kbhit?redirectedfrom=MSDN&view=msvc-160&viewFallbackFrom=vs-2019) _might_ help there? That or [`PeekConsoleInput`](https://docs.microsoft.com/en-us/windows/console/peekconsoleinput). In either case that's for later, and maybe for crossterm itself to integrate it.

arxanas:
Updated. If we commit to Unix-only non-bkocking reads, then we can get rid of the input threqd/channels altogether, which makes it much simpler.

cursive/src/backends/termion.rs
```diff
@@ -40,8 +39,59 @@ pub struct Backend {
...
+///
+/// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+/// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+/// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+/// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+/// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+/// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+/// SOFTWARE.
+#[cfg(unix)]
+fn set_blocking(file: &mut File, blocking: bool) -> std::io::Result<()> {
```
gyscos, line 74: Nit: do we need `&mut` here? `AsRawFd` gives `fn as_raw_fd(&self) -> RawFd`. I think it might work with `&File`. This way you also don't need to make `input_file: File` mut in `init_with_files`.

arxanas: Removed `mut`.

gyscos:
Looks perfect! Thanks again for all that! :D

haydenflinner:
Hi, I guess this patch has been out there a while and has been OK, but it could lead to one of the most annoying bugs I've ever had to debug, so I feel I have to mention it here. 

On many Unix boxes, stdin and stdout are file descriptor**ors** pointing to the same file descrip**tion**. This is because stdout and stderr are just a dup of stdin. Setting nonblocking sets the flag on the file descrip**tion**, which implies two things that are unintuitive. 
One, that making stdin nonblocking also made stdout nonblocking. That may have implications on apps using this crate. 
Two, that closing your program does not restore the blocking flag. So, any subsequent programs which expect stdin/out to be nonblocking will be broken. This has surfaced for me before when `npm` set stdin to nonblocking and then subsequent processes in my shell script always failed, but only if `npm` had been run before them, and *then* only if within a shell script, not at the prompt (because the prompt set stdin to blocking, guarding against this). 

So, we should always make a best effort to set stdin back to blocking before exiting, which I don't see code in this patch to do, but I am new to Rust so I might be missing it. Thanks for reading 😃 

SO Answer on how to undo, if it impacts your terminal -- https://stackoverflow.com/a/48011063
An old writeup / warning -- https://web.archive.org/web/20161023162000/https://jdebp.eu/FGA/dont-set-shared-file-descriptors-to-non-blocking-mode.html

gyscos:
That's a great point!

Just updated the termion backend to set the blocking flag back to true when de-initializing the backend.

PR #583, opened 2021-05-29
Add launchk to showcase

mach-kernel:
This is my first TUI app: always wanted a small launchd tool the same way [some exist for systemd on Linux](https://github.com/linuxenko/chkservice) -- decided to try making one! Thanks to everyone for their work on Cursive, it was very approachable to do!!!


gyscos:
Contrats on that! Looks really nice!

PR #586, opened 2021-06-11
Fix Align::bot_right

jwuensche:
The `Align::bot_right` method created a top right alignment before. This patch changes this to produce
the correct struct variant.

Just found this while writing a few tests, seems to have been a small copy paste error :smile:


gyscos:
Whoops, indeed! Thanks for finding that!

PR #587, opened 2021-06-14
Add `git-branchless` to showcase

arxanas:
This CLI application uses `cursive` for some of the interactive parts, such as `git undo`.

By the way, would you mind releasing a new prerelease version of `cursive`? That way, I can publish my dependent package on https://crates.io.


gyscos:
Hi, and congratulation on the nice tool!

I didn't think of pre-releases, but I certainly can do that! :)

PR #590, opened 2021-07-08
fix the menubar example

chen244:
The menubar example cannot compile , and I fix it
And I am confused that you only have MenuTree function, but in the docs it is Tree, Is it because your doc has not update?


chen244:
I see , the example is in the git..

chen244:
I got a mistake,,

PR #600, opened 2021-08-15
Fix crates.io shield

ColonisationCaptain:
The current source for the icon is offline


gyscos:
Whoops, thanks for the fix!

PR #601, opened 2021-08-17
Remove all 'needless borrows' Clippy complains about

schneiderfelipe:
This removes [`needless_borrow`](https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow) warnings.

### DIFF
--- a/cursive/src/backends/puppet/observed.rs
+++ b/cursive/src/backends/puppet/observed.rs
@@ -281,7 +281,7 @@ pub trait ObservedPieceInterface {
                 match &self.parent()[Vec2::new(x, y)] {
                     None => s.push(' '),
                     Some(cell) => {
-                        if let GraphemePart::Begin(lex) = &cell.letter {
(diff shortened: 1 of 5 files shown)


cursive/src/backends/puppet/observed.rs
```diff
@@ -282,7 +282,7 @@ pub trait ObservedPieceInterface {
                     None => s.push(' '),
                     Some(cell) => {
                         if let GraphemePart::Begin(lex) = &cell.letter {
```
gyscos, line 284: In addition, I'd rather not use implicit reference pattern here.
```suggestion
                        if let GraphemePart::Begin(ref lex) = cell.letter {
```

gyscos (commented):
Thanks for the work! Just a minor detail and it should be good to merge.

cursive/src/backends/puppet/observed.rs
schneiderfelipe: Good catch! Sorry for not seeing the notification early enough! Thanks for the merge!

PR #609, opened 2021-10-16
add wiki-tui to showcases

Builditluc:
I made a simple tui program for wikipedia in cursive. Would you mind adding it to the showcase?


gyscos:
Thanks!

Issue #613, opened 2021-11-06
[BUG] adding content causes views to be pushed off screen.


TwoClocks:
The following code walks the bottom panel off the bottom of the screen.

While I suppose its height is still the required height, but it's off the screen, so who knows.

I think this is a bug?

If it's not, how do I keep both panels on the screen?

Great crate BTW. It's been fun to use.

```rust
fn main() {
    let mut ui = CursiveRunnable::default();
    let mut ui = ui.runner();

    let mut bottom_content = String::with_capacity(1023);
    for n in (1..11).rev() {
        bottom_content.push_str(&format!("{} Line num\n", n));
    }

    ui.add_layer(
        LinearLayout::vertical()
            .child(ResizedView::with_min_height(
                10,
                Panel::new(TextView::new("top content\n").with_name("top")).title("Top Panel"),
            ))
            .child(ResizedView::with_min_height(
                10,
                Panel::new(TextView::new(bottom_content).with_name("bottom")).title("Bottom Panel"),
            )),
    );

    ui.add_global_callback('q', |s| s.quit());

    ui.refresh();
    ui.step();

    while ui.is_running() {
        std::thread::sleep(std::time::Duration::from_millis(500));

        ui.call_on_name("top", |view: &mut TextView| {
            view.append(format!("now is {:?}\n", std::time::Instant::now()))
        });

        ui.refresh();
        ui.step();
    }
}

```

TwoClocks:
Yep. Adding a scrollbar fixes the issue.

I'm not sure what you want to do w/ my layout logic PR, which also fixes the issue w/o needing to add a scroll bar. I'm obviously biased, but I think it's a better general solution. Also easy to add weights too (I think). Let me know if you have any interest... I'd be happy to work on it w/ you. If not, also happy to delete the PR.

PR #614, opened 2021-11-07
cut at new layout logic for LinearLayout

TwoClocks:
This is an attempt to fix [Bug #613](https://github.com/gyscos/cursive/issues/613https://github.com/gyscos/cursive/issues/613) 
In theory, this code could be placed inside the `desperation` branch where `cap` is called. To fix the bug, you have to do most of the work that layout does, so it's here as a replacement for the layout code.

I wouldn't take this PR as is. I am not that familiar w/ cursive, nor rust. It likely needs some tweaking for both.
I think the algorithm is sound and generates similar (but different) results as the previous algorithm (plus bug fix). I personally think it's simpler? but that might be just because I wrote it.

If there is another / better / simpler way to address the bug in question, I'd be happy to take a stab at that w/ some guidance.

### DIFF
--- a/cursive-core/src/views/linear_layout.rs
+++ b/cursive-core/src/views/linear_layout.rs
@@ -505,6 +505,7 @@ impl View for LinearLayout {
         }
     }
 
+    #[allow(unreachable_code)]
     fn required_size(&mut self, req: Vec2) -> Vec2 {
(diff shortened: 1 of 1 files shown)


TwoClocks:
I think it would be fairly easy to update this logic to add weight support.

PR #616, opened 2021-11-12
Add Dialog the capability to be focusable

cecton:
Hello!

I found this feature missing: you can set a dialog to be "focusable". This
means you can focus on the dialog itself, it will take the title primary color
and it can take a callback function.

The callback function is optional: if you don't use the callback, pressing
enter will give the focus to the content of the dialog.

If you press Tab or Shift+Tab it will start focusing the buttons instead.

This is similar to the "interactive" feature of the component
[Card](https://blueprintjs.com/docs/#core/components/card) of Blueprint.js

I will use it to make a new CLI for my library
[gptman](https://github.com/cecton/gptman/).

### DIFF
--- a/cursive-core/src/printer.rs
+++ b/cursive-core/src/printer.rs
@@ -398,6 +398,8 @@ impl<'a, 'b> Printer<'a, 'b> {
     /// If `invert` is `true`, and the theme uses `Outset` borders, then the
     /// box will use an "inset" style instead.
     ///
+    /// If `highlight` is `true`, then the "title" primary color will be used instead.
+    ///
     /// # Examples
     ///
     /// ```rust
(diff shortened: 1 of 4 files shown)


cecton:
@gyscos :grin:

gyscos:
Hi! Yes yes, will take a look soon. A bit busy these days :(

In any case, thanks for the work!

gyscos:
I'm wondering if it could be maybe a better fit for panels?
`Dialog` is a bit of a mix of many things (in big part because it was one of the first ever views to be implemented, so it has quite some baggage...), and is meant to be a convenient "top-level" view in a layer, rather than a real component to embed in a layout.

The more recent `Panel` is trying to be the more minimal and re-usable counterpart. I think making panels focusable could be a great idea!

Do you think it would work for your use-case? In particular, panels don't have the same convenient integrated row of buttons (you would have to add the buttons in a LinearLayout or something).

cursive-core/src/printer.rs
```diff
@@ -407,13 +409,14 @@ impl<'a, 'b> Printer<'a, 'b> {
...
     /// # let printer = Printer::new((6,4), &t, &*b);
-    /// printer.print_box((0, 0), (6, 4), false);
+    /// printer.print_box((0, 0), (6, 4), false, false);
     /// ```
     pub fn print_box<T: Into<Vec2>, S: Into<Vec2>>(
         &self,
         start: T,
         size: S,
         invert: bool,
+        highlight: bool,
```
gyscos, line 419: I think at this point it may be better to use some parameter-struct (to have named parameters) rather than additional parameters:

```rust
use BoxStyle as BS;

printer.print_box(start, size, BS { invert: false, highlight: false });
```

cecton:
Yes I noticed Dialog kinda grew organically...

I will see to move this to panel

PR #617, opened 2021-12-09
Add SACI-RS on the showcase

ihercowitz:
Add another project to serve as an example on how to use Cursive


gyscos:
Looks great, thanks!

Issue #370, opened 2019-08-25
layout using percentages


chrunchyjesus:
Asking for help.

# Problem description

Is it possible to use percentages to layout? For example, I want to have two views, one left and one right, whereas the left one uses 40% of the available width and the right one 60% like in the screenshot below.

I have searched solutions on doc.rs and issues on github, but did not find anything relevant, thus this question.

![image](https://user-images.githubusercontent.com/20382692/63650281-749c3100-c738-11e9-8440-88fefd45a228.png)

chrunchyjesus:
Ah, okay. So that's why the weight parameters did nothing when I changed them. Thanks for the info! :+1:

jaygazula27:
@gyscos In the meantime, do you have any recommendations on how we can achieve this sort of layout even if it's not necessarily easy?  
The only thing I can come up with is somehow calculate the total width of screen and then adjust each of the (left and right) panels' width accordingly during a resize event. But I'm approaching it from a web UI point of view, and I'm not sure if that can be translated in a TUI.

gyscos:
One way to achieve it is to basically re-implement a simplified version of `LinearLayout` which ignores the children's preferences and just give them a percentage of the available space.

If you want to use fixed percentages, you might be also be able to implement this using [`FixedLayout`](https://docs.rs/cursive_core/0.1.1/cursive_core/views/struct.FixedLayout.html) wrapped in an [`OnLayoutView`](https://docs.rs/cursive_core/0.1.1/cursive_core/views/struct.OnLayoutView.html): on each call to `layout`, update the children position in the `FixedLayout` using the percentages.

jaygazula27:
@gyscos Thanks for that information! I'll try to follow what you said and figure out a decent solution. Appreciate the help.

llambdaa:
What is the state of this feature? Are you planning to implement it in the near future, so does it have some priority?
Having relative layouts using ratios would be a major feature and probably very appreciated!

I figure, for that, you must be able to enforce width and height onto components that are respected when drawing it so that e.g. a button actually renders with the given width instead of shrinking to the bare minimum it needs. Resolving the ratios inside a LinearLayout and calculating the resulting width and or height (that could then be given as a requirement to the child components) shouldn't be that much of an issue.

PR #618, opened 2021-12-12
Clarify that weight is currently unused by LinearLayout

Keating950:
According to [this comment](https://github.com/gyscos/cursive/issues/370#issuecomment-524979221) from a few years back on #370, the weight values for children of `LinearLayout` currently aren't used in the layout process, and looking at the source code, that appears to still be true. In this case, I think it would be helpful to note that in the method docs.

### DIFF
--- a/cursive-core/src/views/linear_layout.rs
+++ b/cursive-core/src/views/linear_layout.rs
@@ -147,7 +147,7 @@ impl LinearLayout {
         }
     }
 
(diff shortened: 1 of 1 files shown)


gyscos:
Thanks! Yeah I always felt like "but I'm juuuust about to actually implement it". I guess it's time I accept it will take a bit longer to actually implement.

PR #622, opened 2022-01-02
Add showcase application

DevHyperCoder:
Add [rbmenu-tui](https://github.com/DevHyperCoder/rbmenu-tui) to the list of showcase application list.

It is a easy to use bookmark manager

Thanks!


gyscos:
Congratulation! Looks like a great tool!

PR #657, opened 2022-06-20
Add to_string() method for events

Gip-Gip:
Went ahead and added to_string methods for events, I find them very useful for debugging


gyscos:
Note that events should already implement Debug - is that not enough for your use case?

Gip-Gip:
> Note that events should already implement Debug - is that not enough for your use case?

Just learned about debug, sorry learning the language as I go lol

PR #658, opened 2022-06-30
scroll: in strict mode, fill non-scrolling axis

Mstrodl:
Non-scrolling axes wouldn't expand to fill their parent otherwise.

See also: https://github.com/ComputerScienceHouse/clink/pull/18

Without this change, the `SelectView` I put in my `Dialog` doesn't expand to fill the much wider `Dialog`


gyscos:
Thanks for the PR!

Indeed, looks like a bug. :fearful:

Mstrodl:
> Thanks for the PR!
> 
> Indeed, looks like a bug. fearful

Thank you! Any idea when this will release?

Mstrodl:
> Any idea when this will release?

Nevermind, you're very fast! :smile:

PR #661, opened 2022-07-10
Add filters example

DJDuque:
I think it would be good to have an example with a similar behavior to my PR: A settings/configuration "pop-up" that allows us to modify `user_data` to be used elsewhere in our application. I guess this is a common-ish thing for people to do, but maybe a little bit cumbersome for beginners like myself.

With that said, this is mostly a request for someone with more experienced to have a look at my PR and hopefully make it simpler, less repetitive, etc. I think many people could benefit from more examples with "complex" but common patterns.


cursive/examples/filters.rs
```diff
@@ -0,0 +1,135 @@
...
+                                    )
+                                    .child({
+                                        let mut button = color_group
+                                            .button(Some(Color::Red), "Red");
+                                        if let Some(Color::Red) = current_color
+                                        {
+                                            button.select();
+                                        }
+                                        button.fixed_width(10)
+                                    })
```
gyscos, line 67: Nit: you can use [`With`](https://docs.rs/cursive/latest/cursive/trait.With.html) for that:
```suggestion
                                    .child(
                                        color_group
                                            .button(Some(Color::Red), "Red")
                                            .with_if(current_color == Some(Color::Red), RadioButton::select)
                                            .fixed_width(10))
                                    )
```

cursive/examples/filters.rs
```diff
@@ -0,0 +1,135 @@
...
+
+fn main() {
+    let mut siv = cursive::default();
+    siv.set_user_data(Filters::default());
+    // Leave the menu visible
+    siv.set_autohide_menu(false);
+    siv.menubar()
+        .add_leaf("Filters", |s| {
+            // Hide menu while "pop-up" settings are visible
+            s.set_autohide_menu(true);
```
gyscos, line 33: I think this is more up to personal preference, and could be removed from the example to keep it more focused.

cursive/examples/filters.rs
```diff
@@ -0,0 +1,135 @@
...
+
+            let mut color_group: RadioGroup<Option<Color>> = RadioGroup::new();
+            let mut size_group: RadioGroup<Option<Size>> = RadioGroup::new();
+
+            // Get current filters to draw the correct selected buttons.
+            let user_data = s.user_data::<Filters>().unwrap();
+            let current_color = user_data.color;
+            let current_size = user_data.size;
```
gyscos, line 41: Note that `.user_data` keeps `Cursive` borrowed. While rust is smart enough to "drop" the borrow here just in time to let us call `.add_layer`, it may be safer to use something like `.with_user_data` instead to more clearly limit the borrow:

```suggestion
            let (current_color, current_size) = s.with_user_data(|filters: &mut Filters| {
                (filters.color, filters.size)
            };
```

cursive/examples/filters.rs
```diff
@@ -0,0 +1,135 @@
...
+                                        button.fixed_width(10)
+                                    }),
+                            ),
+                    )
+                    .button("Done", move |s| {
+                        // Save selected filters as user_data
+                        s.user_data::<Filters>().unwrap().color =
+                            *color_group.selection();
+                        s.user_data::<Filters>().unwrap().size =
+                            *size_group.selection();
```
gyscos, line 120: Here too, `with_user_data` may help:

```suggestion
                        s.with_user_data(|filters: &mut Filters| {
                            filters.color = *color_group.selection();
                            filters.size = *size_group.selection();
                        }).unwrap(); 
```

cursive/examples/filters.rs
```diff
@@ -0,0 +1,135 @@
...
+
+            // Get current filters to draw the correct selected buttons.
+            let user_data = s.user_data::<Filters>().unwrap();
+            let current_color = user_data.color;
+            let current_size = user_data.size;
+            s.add_layer(
+                Dialog::new()
+                    .title("Filters")
+                    .content(
+                        LinearLayout::vertical()
```
gyscos, line 46: You can use `ListView` for this: it's a vertical list of views with labels.

https://github.com/gyscos/cursive/blob/main/cursive/examples/list_view.rs

Could probably be renamed `FormView` to be more obvious about the intent.

gyscos (commented):
Thanks for the work!

So the main idea of this example is to add a more complex use of user_data to prepare some state and use it. It's probably needed; `user_data` was added "relatively" recently, and may not have enough examples showing how to use it.

cursive/examples/filters.rs
DJDuque: `RadioButton::select` returns `EventResult`; so I had to change it for `|button| {button.select(); }`

cursive/examples/filters.rs
DJDuque: I left this in because otherwise we could keep opening "Filter" windows on top of each other. I agree it has nothing to do with the example; it just looks nicer. Let me know if you still want me to remove it.

cursive/examples/filters.rs
```diff
@@ -0,0 +1,138 @@
...
+                                        color_group
+                                            .button(Some(Color::Blue), "Blue")
+                                            .with_if(
+                                                current_filters.color
+                                                    == Some(Color::Blue),
+                                                |button| {
+                                                    button.select();
+                                                },
+                                            ),
+                                    ),
```
gyscos, line 80: You could also set the focus of the LinearLayout itself:
```suggestion
                                    )
                                    .with(|layout| {
                                        layout.set_focus_index(match current_filters.color {
                                            None => 0,
                                            Some(Color::Red) => 1,
                                            Some(Color::Blue) => 2,
                                            _ => unreachable!(),
                                        }).unwrap();
                                    }),
```

gyscos (commented):
As I mentioned in another comment it may be simpler to just open this new popup from a button in the first dialog, and not have a menubar at all.

As for some of the repetition, what might help is a helper to build a view from some data - for example build the radio buttons from a few values of a given type (in our case an `Option<Color>`). Maybe even some way to derive the entire view from the `Filters` struct itself - though this is a much larger project, clearly outside the scope of this example.

gyscos:
Was having fun trying to simplify the way the radio buttons are generated:

```rust
fn make_radio<T: 'static + PartialEq>(                              
    values: impl IntoIterator<Item = (impl Into<String>, T, usize)>,
    group: &mut RadioGroup<T>,                                      
    current_value: &T,                                              
) -> impl cursive::View {                                           
    LinearLayout::horizontal().with(|layout| {                      
        for (label, value, width) in values.into_iter() {           
            let selected = &value == current_value;                 
            layout.add_child(                                       
                group                                               
                    .button(value, label)                           
                    .with_if(selected, |b| {                        
                        b.select();                                 
                    })                                              
                    .fixed_width(width),                            
            );                                                      
            if selected {                                           
                layout.set_focus_index(layout.len() - 1).unwrap();  
            }                                                       
        }                                                           
    })                                                              
}                                                                   

// ...

ListView::new()                                  
    .child(                                      
        "Color: ",                               
        make_radio(                              
            [                                    
                ("Any", None, 10),               
                ("Red", Some(Color::Red), 10),   
                ("Blue", Some(Color::Blue), 10), 
            ],                                   
            &mut color_group,                    
            &current_filters.color,              
        ),                                       
    )                                            
    .child(                                      
        "Size: ",                                
        make_radio(                              
            [                                    
                ("Any", None, 10),               
                ("Small", Some(Size::Small), 10),
                ("Large", Some(Size::Large), 10),
            ],                                   
            &mut size_group,                     
            &current_filters.size,               
        ),                                       
    ),                                           
```

Not sure it's actually simpler, but it's a bit less copy/pasted.

DJDuque:
> (First, thank you again for the continued work here!)

On the contrary, thanks a lot for the constant reviews; it is very helpful.

I have pushed some changes based on your last comments, please let me know if you have any more suggestions.

gyscos:
Thanks for the work!
I don't have a lot of time for more review so I just merged it, we can always improve the example later if we need :)

Thanks again!

PR #665, opened 2022-07-22
select_view: Cache required_size

brianc118:
Cache required size when we know it doesn't need to be recomputed. Cache
is invalidated whenever we take mutable reference to items or mutate
items.

https://github.com/facebookincubator/below has an issue where the UI is
very slow because we have >1000 processes in the process view. This
patch fixes it.

Tested with `cargo test`

### DIFF
--- a/cursive-core/src/views/select_view.rs
+++ b/cursive-core/src/views/select_view.rs
@@ -73,6 +73,9 @@ pub struct SelectView<T = String> {
     // We "cache" it during the draw, so we need interior mutability.
     last_offset: Cell<Vec2>,
     last_size: Vec2,
+
+    // Cache of required_size. Set to None when it needs to be recomputed.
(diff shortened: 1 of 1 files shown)


gyscos:
Looks good, thanks for the work!

brianc118:
Thanks, might we expect to see a version bump?

gyscos:
cursive-core 0.3.4 was released with this fix (and another fix for FixedLayout).

Separately, cursive 0.19.0 was also released to update crossterm to 0.24.0.

(Both changes should be independent: you can use cursive-core 0.3.4 with cursive 0.18.0 or cursive-core 0.3.3 with cursive 0.19.0.)

brianc118:
Hmm I don't see cursive 0.19.0 but I do see cursive-core 0.3.4 on crate.io

PR #669, opened 2022-08-14
updated the documentation for call_on_name

alexanderkjall:


cursive-core/src/cursive.rs
```diff
@@ -521,9 +521,9 @@ impl Cursive {
         self.root.call_on(sel, callback)
     }
 
-    /// Tries to find the view identified by the given id.
+    /// Tries to find the view identified by the given name.
     ///
-    /// Convenient method to use `call_on` with a `view::Selector::Id`.
+    /// Convenient method to use `call_on` with a `view::Selector::Name`.
```
Apitronix, line 526: Prefer mention "view::Selector::Name" as an internal link than a simple string.
You can do it using square brackets to surround the backwards simple quotes around your string. And don't forget to build the doc to ensure everything works before committing.

Apitronix (commented):
Thank you for your update, there is a little thing you can do to improve it again. :)

cursive-core/src/cursive.rs
alexanderkjall: Thanks for the suggestion, now fixed :)

gyscos (approved):

gyscos:
Thanks for the work!

PR #673, opened 2022-08-22
Improve Printer documentation

Apitronix:


cursive-core/src/printer.rs
```diff
@@ -18,10 +19,9 @@ use unicode_width::UnicodeWidthStr;
 
 /// Convenient interface to draw on a subset of the screen.
 ///
-/// The area it can print on is defined by `offset` and `size`.
-///
-/// The part of the content it will print is defined by `content_offset`
-/// and `size`.
+/// The area it can print on is defined by `offset` and `size`.\
```
gyscos, line 22: Why this change?

[The doc format](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html) (much like the git message format) relies on a single short line, followed by a blank line, followed by the rest of the documentation.

cursive-core/src/printer.rs
```diff
@@ -118,16 +115,53 @@ impl<'a, 'b> Printer<'a, 'b> {
         }
     }
 
+    /// Prints some text at the given position.
+    ///
+    /// # Parameters
+    /// `start` is an offset used to print the text in the view.\
+    /// `text` is a simple string to print on the screen.
```
gyscos, line 122:
```suggestion
    /// * `start` is the offset of the first character to print.
    /// * `text` is a simple string to print on the screen.
```

cursive-core/src/printer.rs
```diff
@@ -118,16 +115,53 @@ impl<'a, 'b> Printer<'a, 'b> {
...
+    ///     pub fn new() -> Self {
+    ///         Self {
+    ///             word: String::from("Eh, tu connais Rust?"),
+    ///         }
+    ///     }
+    /// }
+    /// 
+    /// impl View for CustomView {
+    ///     fn draw(&self, printer: &Printer<'_, '_>){
+    ///         printer.print(XY::new(0, 0), self.word.as_str());
```
gyscos, line 146:
```suggestion
    ///         printer.print(XY::new(0, 0), &self.word);
```

I think the example could be reduced to only include a `draw` function, to focus on the `print` method (and not on implementing a view).

gyscos (commented):
Hi, and thanks for the work!

I added a few comments.
Mostly I'm not sure what we gain by replacing newlines with backslashes.

cursive-core/src/printer.rs
```diff
@@ -1,3 +1,4 @@
+#![warn(missing_docs)]
```
gyscos, line 1: I don't think we need this, since it's already set at the crate level.

Apitronix: Ok, then I will remove it. :)

cursive-core/src/printer.rs
Apitronix: Using less words to say the same thing improve the readability of the sentence. Here the goal is not to loose the reader using few unnecessary words.

Furthermore, you have the short line (line 20), and then the rest of the documentation (line 22 to line 24).

The goal of the blank line is to create few paragraphs. If you use the "\" character, then you add a carriage return *inside* your paragraph. So you need to use the blank line if you want to create another paragraph, and use the "\" character if you want to structure a single paragraph. That's why I didn't add the "\" char at the end of each line.

cursive-core/src/printer.rs
Apitronix: The first char is not the only one affected by the `start` parameter. This parameter affect the whole `text`.

cursive-core/src/printer.rs
Apitronix: If I want to implement the `required_size` method here, it's because if you are new with your library, then you could be in trouble to understand why your call to the `print` method leads to a bit of an "unexpected behavior" (by cutting the `text` before its end).

gyscos: I agree, but this is a concern of implementing your own view (which could be part of the View documentation). I don't think we should mention this in _every_ example for the printer.

As for `&` vs `.as_str`, it's _quite_ idiomatic to use the `Deref<Target=str>` from `String`, I don't think anyone would be surprised.

cursive-core/src/printer.rs
gyscos: Ah indeed, I thought this was the first line of the doc comment.

Since it's just to merge the two paragraphs into a single one with newline?

cursive-core/src/printer.rs
Apitronix: Ok, so I will remove the `as_str` call. Maybe I can warn in the global Printer class documentation, about the behavior of the printer on the views?

gyscos: This is really something that belongs in the view - you only care about that when you implement your own view.

We _can_ mention in the doc that anything further than the printer's `size` would be cropped out.

cursive-core/src/printer.rs
Apitronix: Yes, that's it. It makes the doc more easier to read than having a new paragraph for each sentence. :smile:

cursive-core/src/printer.rs
Apitronix: Ok, I've done the modification on my last commits.

It could be good to mention that in the `Printer` global documentation too, maybe (`Printer` not strictly related to `View`). And maybe that's why Printer's members stay a bit... obscure to me.

Apitronix: As I understand what you are saying, `Printer` is always related to one view, but a single view can "contains" one or more printers.
Do you have an example of a `Printer` that works without any view?

cursive-core/src/printer.rs
```diff
@@ -93,8 +94,7 @@ impl<'a, 'b> Printer<'a, 'b> {
 
     /// Clear the screen.
     ///
-    /// It will discard anything drawn before.
-    ///
+    /// It will discard anything drawn before.\
```
gyscos, line 97: In this case I think it is a separate piece of information that deserves its own paragraph - it's unrelated to "discarding anything drawn before".

cursive-core/src/printer.rs
```diff
@@ -46,8 +48,7 @@ pub struct Printer<'a, 'b> {
     /// Offset into the view for this printer.
     ///
     /// The view being drawn can ignore this, but anything to the top-left of
-    /// this will actually be ignored, so it can be used to skip this part.
-    ///
+    /// this will actually be ignored, so it can be used to skip this part.\
```
gyscos, line 51: Here too I think the formula for the actual location is independent from "the view can ignore this", so I don't think it belongs in the same paragraph.

cursive-core/src/printer.rs
```diff
@@ -118,16 +118,47 @@ impl<'a, 'b> Printer<'a, 'b> {
         }
     }
 
+    /// Prints some text at the given position.
+    ///
+    /// # Parameters
+    /// * `start` is the offset used to print the text in the view.
+    /// * `text` is a simple string to print on the screen.
```
gyscos, line 125: We can also specify that `text` should really be a single line, no line-wrapping will be done here.

cursive-core/src/printer.rs
Apitronix: @gyscos Am I right ?
Do you ant me to add this to the documentation, or do you want to do it yourself ?

gyscos: In practice `Printer` is always "associated" with a view, though this association is merely implicit (it just happens to be given to the view in its `draw` function).

The _concept_ of Printer doesn't _need_ views. For example for some time the menubar was not really a `View`, but was still using the `Printer`. Same for the background. Today these have been made into views, but that's more a detail.

I don't think it's very important though. We may add doc about it later if it becomes useful.

gyscos (approved):
Thanks! And sorry for the many rounds of review. I never used these trailing backslashes, so it took some time to get used to them. :upside_down_face:

Apitronix:
Not a problem ! I prefer working with people who take care of their project by taking the time needed to make a good review. :wink:

PR #676, opened 2022-09-19
Fix for immut3 macro

wigy-opensource-developer:
It contained a copy-paste typo. So I have fixed it and added a test for it. I also had to fix 2 doc-test on the way.


gyscos (approved):
Whoops indeed, thanks!

PR #679, opened 2022-09-27
Replace term_size with terminal_size

dbrgn:
`term_size` is unmaintained, see https://rustsec.org/advisories/RUSTSEC-2020-0163.

(On that note, is the dependency actually used somewhere? I can't find any usages... Maybe we can remove it completely?)


gyscos:
Hi, and thanks for the PR!

Indeed, it seems we stopped using term_size some time ago, so let's just remove it, no need for `terminal_size` :)

cursive/Cargo.toml
```diff
@@ -22,7 +22,7 @@ unicode-segmentation = "1"
 unicode-width = "0.1"
 lazy_static = "1"
 libc = "0.2"
-term_size = { version = "0.3", optional = true }
+terminal_size = { version = "0.2", optional = true }
```
gyscos, line 25:
```suggestion
```

cursive/Cargo.toml
```diff
@@ -53,8 +53,8 @@ version = "0.25"
 doc-cfg = ["cursive_core/doc-cfg"]  # Enable doc_cfg, a nightly-only doc feature.
 blt-backend = ["bear-lib-terminal"]
 default = ["ncurses-backend"]
-ncurses-backend = ["ncurses", "maplit", "term_size"]
-pancurses-backend = ["pancurses", "maplit", "term_size"]
+ncurses-backend = ["ncurses", "maplit", "terminal_size"]
+pancurses-backend = ["pancurses", "maplit", "terminal_size"]
```
gyscos, line 57:
```suggestion
ncurses-backend = ["ncurses", "maplit"]
pancurses-backend = ["pancurses", "maplit"]
```

gyscos (approved):

PR #686, opened 2022-10-21
Add `invert` method to `ColorStyle`

linclelinkpart5:
I added a simple method in my `cursive`-using project to do this, but I figured it would be helpful to add this upstream as a method on `ColorStyle`. Followed the same pattern as `ColorPair::invert`.


gyscos (approved):

gyscos:
Looks great, thanks for the work!

PR #687, opened 2022-10-21
Add example of status bar extension trait

joelparkerhenderson:
This example demonstrates:

1. How to create a status bar line at the bottom of the screen by using a StatusBarExt trait. The trait encapsulates the ideas in the existing example file `status.rs`.

2. More generally, how to create an extension trait that attaches new functionality to an existing cursive struct. The  trait enables functions to be called on the root, such as `siv.status_bar("Hello World")` which creates a status bar with the given content.


gyscos:
Thanks! More examples is always a good thing :)

PR #691, opened 2022-11-12
checkbox: remove missing method from example

rrooij:
The with_name method is not part of the checkbox view and therefore, the provided example does not work.

### DIFF
--- a/cursive-core/src/views/checkbox.rs
+++ b/cursive-core/src/views/checkbox.rs
@@ -17,7 +17,7 @@ type Callback = dyn Fn(&mut Cursive, bool);
 /// use cursive_core::traits::Nameable;
 /// use cursive_core::views::Checkbox;
 ///
-/// let checkbox = Checkbox::new().checked().with_name("check");
+/// let checkbox = Checkbox::new().checked();
 /// ```
 pub struct Checkbox {
     checked: bool,


gyscos:
Do you mean the example does not currently compile?

If you run `cargo test --doc checkbox` you will see that it compile and correctly run this test.

The `with_name` method comes from the [`Nameable`](https://docs.rs/cursive/latest/cursive/traits/trait.Nameable.html) trait imported just above.

I agree it's not _strictly_ required to use a checkbox, but in practice users will often want to give this view a name to retrieve its state, so I think it's fine to include it here, to "look (a bit) more like a real-life example".

rrooij:
Hah, that you are totally right. For some reason, my Rust compiler was complaining about `with_name` missing and I found out why. It is because the examples use `cursive_core` instead of `cursive`. Which wasn't imported in my case. Wouldn't it be better if the examples use `cursive` instead? I'm closing this PR as you are right that it clearly implements the Nameable trait, but the example should perhaps use `cursive` instead of the imports by `cursive_core`.

gyscos:
Indeed, I just changed it to use `cursive`:
https://github.com/gyscos/cursive/blob/main/cursive-core/src/views/checkbox.rs#L18-L19

PR #696, opened 2022-11-23
Reset focus field when clearing LinearLayout

cdxjcl:
The focus field should be reset to the default value when clearing the LinearLayout view.


gyscos:
Good point, thanks!

gyscos (approved):

PR #699, opened 2022-12-11
Update tutorial_1.md with `use cursive;`

oliverralbertini:
Text says to import cursive, but just has an empty main().


oliverralbertini:
Closing, since the import is not needed to start using e.g. `cursive::default()`.

PR #716, opened 2023-02-22
Let crossterm to initialyze with file for stdout

mikem-zed:
Currently crossterm chooses /dev/tty by default for *nix however it may be required to initialize it with e.g. /dev/stdout

it can be used like this 

```Rust        
let cur = cursive::CursiveRunnable::new(|| {
            cursive::backends::crossterm::Backend::init_with_stdout_file(std::fs::File::create(
                "/dev/stdout",
            )?)
        })
```

### DIFF
--- a/cursive/src/backends/crossterm.rs
+++ b/cursive/src/backends/crossterm.rs
@@ -10,6 +10,9 @@ use std::{
     time::Duration,
 };
 
+#[cfg(unix)]
+use std::fs::File;
+
 pub use crossterm;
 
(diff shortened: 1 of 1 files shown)


gyscos:
Sorry for the delay - this looks great! Thanks!

gamma-delta:
Do we have an ETA on when this is going to be put into an official release? I'm working on a crate that lets you control two terminals with one Cursive application (the normal terminal and another tty) but it needs this MR to be up on crates.io

Issue #714, opened 2023-02-10
[BUG] Log flood in the debug console by default.


vi:
**Describe the bug**

Debug console with logs show too many `DEBUG`-level messages by default.

**To Reproduce**

Implement simple app based on README and some functions' documentation. In particular, use `toggle_debug_console` and `logger::init`.

```rust
    let mut siv = Cursive::new();
    cursive::logger::init();

    let linear_layout = LinearLayout::vertical()
    .child(TextView::new("Top of the page"))
    .child(TextArea::new().fixed_size((20, 5)))
    .child(Button::new("Ok", |s| s.quit()));
    siv.set_window_title("myapp");
    siv.add_fullscreen_layer(linear_layout);
    
    siv.add_global_callback('~', Cursive::toggle_debug_console);
    siv.add_global_callback('q', |s| s.quit());
    
    siv.run();
```

**Expected behavior**

Resulting console activated by `~` contains few logs (unless errors happen) from Cursive itself (e.g. just something like "Hello from Cursive version ... running on backend ... on terminal ... on locale ...", "Main window initialized", etc.). New logs from user app are easily visible in it. Interaction with the debug console itself (e.g. scrolling it) does not typically produce new logs. Debug-level logs (especially from Cursive itself) appear only if user explictily configures it from within the code or environment.

**Actual behavior**

The console is flooded by `DEBUG`-level logs originating from Cursive itself, which pile up more and more as I attempt to scroll the console to the bottom. Even starting with `RUST_LOG=info` does not seems to stop this.

**Environment**
* Operating system used: Linux Debian Bookworm amd64
* Backend used: ncurses
* Current locale: C.UTF-8
* Cursive version: 0.20.0

gyscos:
Hi,
That's interesting; what kind of debug logs are you seeing?

Grepping for `log::` in the repo, I see a couple `warn!`, but that's about it.

The current logger implementation is also very basic, and doesn't offer much filtering. Something like [flexi-logger](https://github.com/deinstapel/cursive-flexi-logger-view) should allow you to specify per-module levels, so you can only allow warnings and errors from cursive, but see info logs from your own module.

blaisdellma:
Running the above example, I'm seeing logs from 

- [LinearLayout::required_size()](https://github.com/gyscos/cursive/blob/1a0cc868b41232c0d5290a11b4b987ffed757798/cursive-core/src/views/linear_layout.rs#L489).
[...]

PR #719, opened 2023-03-06
Add basic features to CursiveLogger

blaisdellma:
Hello,

This is my attempt at following the first of my suggestions in #714 . 

I tried to leave as much alone as I could so that there would be no breaking changes with as many useful features as possible. Anyone using the logger now shouldn't have to change anything and get the exact same behavior. There shouldn't be any breaking changes to the existing API, either in function signatures or behavior.

Logging would use two different filter levels, one each for logs generated internally by cursive and externally by whatever else. This solves the issue in #714 as the debug messages won't clog the log with all the redraws from scrolling (assuming the filter levels are set appropriately). These levels are set either directly or by the environment variables `RUST_LOG` and `CURSIVE_LOG`. I also added the ability to change the log queue size too. I hope this meets option 2 in #467 as a "good enough" feature set.

Two small examples:
```
cursive::logger::CursiveLogger::new()
    .with_int_log_level(log::LevelFilter::Warn)
    .with_ext_log_level(log::LevelFilter::Debug)
    .with_log_size(10000)
    .init();
```

```
cursive::logger::CursiveLogger::new()
    .with_env()
    .with_log_size(10000)
    .init();
```

The whole `Box::leak(Box::new(self))` thing is to avoid having to add the `std` feature flag to `log` and change `Cargo.toml`. This is exactly how `log` handles `set_boxed_logger` internally.

### DIFF
--- a/cursive-core/src/logger.rs
+++ b/cursive-core/src/logger.rs
@@ -3,6 +3,7 @@
 use lazy_static::lazy_static;
 use std::collections::VecDeque;
 use std::sync::Mutex;
+use std::str::FromStr;
 
 /// Saves all log records in a global deque.
 ///
@@ -47,23 +48,6 @@ pub struct CursiveLogger {
     log_size: usize
 }
 
-fn get_env_log_level(env_var_name: &str) -> Option<log::LevelFilter> {
-    match std::env::var(env_var_name) {
-        Ok(mut log_level_str) => {
-            log_level_str.make_ascii_uppercase();
-            match log_level_str {
-                level if level == "TRACE" => Some(log::LevelFilter::Trace),
-                level if level == "DEBUG" => Some(log::LevelFilter::Debug),
-                level if level == "INFO" => Some(log::LevelFilter::Info),
-                level if level == "WARN" => Some(log::LevelFilter::Warn),
-                level if level == "ERROR" => Some(log::LevelFilter::Error),
-                _ => None,
-            }
-        }
-        Err(_) => None,
-    }
-}
-
 impl CursiveLogger {
     /// Creates a new CursiveLogger with default log filter levels of `log::LevelFilter::Trace`.
     /// Remember to call `init()` to install with `log` backend.
@@ -92,12 +76,16 @@ impl CursiveLogger {
     /// If `CURSIVE_LOG` is set, then the internal log level is set to match with precedence over
     /// `RUST_LOG`.
     pub fn with_env(mut self) -> Self {
-        if let Some(filter_level) = get_env_log_level("RUST_LOG") {
-            self.int_filter_level = filter_level;
-            self.ext_filter_level = filter_level;
+        if let Ok(rust_log) = std::env::var("RUST_LOG") {
+            if let Ok(filter_level) = log::LevelFilter::from_str(&rust_log) {
+                self.int_filter_level = filter_level;
+                self.ext_filter_level = filter_level;
+            }
         }
-        if let Some(filter_level) = get_env_log_level("CURSIVE_LOG") {
-            self.int_filter_level = filter_level;
+        if let Ok(cursive_log) = std::env::var("CURSIVE_LOG") {
+            if let Ok(filter_level) = log::LevelFilter::from_str(&cursive_log) {
+                self.int_filter_level = filter_level;
+            }
         }
         self
     }
--- a/cursive-core/src/logger.rs
+++ b/cursive-core/src/logger.rs
@@ -33,6 +33,12 @@ use std::sync::{Mutex, RwLock};
 pub struct CursiveLogger;
 
 lazy_static! {
+    /// Circular buffer for logs. Use it to implement [`DebugView`].
+    ///
+    /// [`DebugView`]: ../views/struct.DebugView.html
+    pub static ref LOGS: Mutex<VecDeque<Record>> =
+        Mutex::new(VecDeque::new());
+
     // Log filter level for log messages from within cursive
     static ref INT_FILTER_LEVEL: RwLock<log::LevelFilter> = RwLock::new(log::LevelFilter::Trace);
     // Log filter level for log messages from sources outside of cursive
@@ -75,14 +81,6 @@ pub struct Record {
     pub time: time::OffsetDateTime,
     /// Message content
     pub message: String,
-}
-
-lazy_static! {
-    /// Circular buffer for logs. Use it to implement [`DebugView`].
-    ///
-    /// [`DebugView`]: ../views/struct.DebugView.html
-    pub static ref LOGS: Mutex<VecDeque<Record>> =
-        Mutex::new(VecDeque::new());
 }
 
 /// Log a record in cursive's log queue.
--- a/cursive-core/src/logger.rs
+++ b/cursive-core/src/logger.rs
@@ -16,7 +16,7 @@ use std::sync::{Mutex, RwLock};
 ///
 /// ```
 /// # use cursive_core::*;
-/// logger::set_filter_levels_with_env();
+/// logger::set_filter_levels_from_env();
 /// logger::init();
 /// ```
 ///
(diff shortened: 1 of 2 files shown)


cursive-core/src/logger.rs
```diff
@@ -2,14 +2,107 @@
...
+    pub fn with_log_size(mut self, log_size: usize) -> Self {
+        self.log_size = log_size;
+        self
+    }
+
+    /// Installs the logger with log. Calling twice will panic.
+    pub fn init(self) {
+        reserve_logs(self.log_size);
+        log::set_logger(Box::leak(Box::new(self))).unwrap();
+        log::set_max_level(log::LevelFilter::Trace);
```
gyscos, line 103: We could set the maximum of (int, ext) levels to avoid logging trace if it wouldn't be seen anyway.

cursive-core/src/logger.rs
```diff
@@ -2,14 +2,107 @@
...
+    /// Sets the size of the log queue
+    pub fn with_log_size(mut self, log_size: usize) -> Self {
+        self.log_size = log_size;
+        self
+    }
+
+    /// Installs the logger with log. Calling twice will panic.
+    pub fn init(self) {
+        reserve_logs(self.log_size);
+        log::set_logger(Box::leak(Box::new(self))).unwrap();
```
gyscos, line 102: We may want to use `set_boxed_logger`, which hides the leaking away so it feels less dirty (even though it's really the same):
https://docs.rs/log/latest/log/fn.set_boxed_logger.html

I guess a "leak-free" version would keep CursiveLogger zero-sized and rely on a global/singleton state instead, but that may be more complex, with quite limited benefits (prevents leaks if `init()` is called repeatedly).

cursive-core/src/logger.rs
```diff
@@ -44,12 +137,18 @@ pub fn log(record: &log::Record) {
 }
 
 impl log::Log for CursiveLogger {
-    fn enabled(&self, _metadata: &log::Metadata) -> bool {
-        true
+    fn enabled(&self, metadata: &log::Metadata) -> bool {
+        if metadata.target().contains("cursive_core") {
```
gyscos, line 141: Maybe we should be more restrictive, requiring this to be a prefix of the target?

gyscos:
Thanks for the work! Just a couple nitpicks.

cursive-core/src/logger.rs
blaisdellma: Done.

cursive-core/src/logger.rs
blaisdellma: Done.

cursive-core/src/logger.rs
blaisdellma: Yeah, I originally had used `set_boxed_logger` but that's hidden behind the `std` feature flag for `log` and I didn't want to change the dependencies unless I had to. 

As written, calling `init()` multiple times panics anyways, so I don't think we have to worry about leaks. Having everything kept in global state might not be a bad idea since we shouldn't expect anyone to create multiple `CursiveLogger` objects anyways.

I made another branch for this: [cursivelogger_global](https://github.com/blaisdellma/cursive/blob/8010b2c071455f5ffab97d485b62833a94a680e0/cursive-core/src/logger.rs). It is a bit more complicated and I don't know if I like the API ergonomics. One plus though is being able to change the log filter levels after initialization. I don't know if that's worth it.

gyscos: Why the `Mutex<Cell<...>>` in that branch? Could a simple mutex work too?

gyscos: I think I like this global branch! No feature change for the log dependency, uses lazy_static that we were already using...

blaisdellma: Yeah, both work, you're right. I don't know what I was thinking with `Mutex<Cell<...>>`. Since `LOGS` is a `Mutex` anyway I'm not sure there's much of a speed difference.

blaisdellma: I created a [PR for cursivelogger_global](https://github.com/blaisdellma/cursive/pull/1). There is a tiny issue with `set_log_size` though.

blaisdellma:
@gyscos
What do you think about the current state with the global logger?
What about the change to not having a `set_log_size()` function and just using `reserve_logs()`? ([for reference](https://github.com/blaisdellma/cursive/pull/1#issuecomment-1461160889))
Are there other changes/issues in the way of merging this?

cursive-core/src/logger.rs
```diff
@@ -1,15 +1,71 @@
...
-static LOGGER: CursiveLogger = CursiveLogger;
+lazy_static! {
+    // Log filter level for log messages from within cursive
+    static ref INT_FILTER_LEVEL: RwLock<log::LevelFilter> = RwLock::new(log::LevelFilter::Trace);
+    // Log filter level for log messages from sources outside of cursive
+    static ref EXT_FILTER_LEVEL: RwLock<log::LevelFilter> = RwLock::new(log::LevelFilter::Trace);
+}
+
+/// Sets the internal log filter level.
+pub fn set_int_filter_level(level: log::LevelFilter) {
```
gyscos, line 43: Nit: could we call it `set_internal_filter_level`, to prevent any ambiguity around `int`?

Let's also extend `set_external_filter_level` for consistency.

I don't care as much about the lazy static global as it's not public.

cursive-core/src/logger.rs
```diff
@@ -1,15 +1,71 @@
...
+/// Sets the external log filter level.
+pub fn set_ext_filter_level(level: log::LevelFilter) {
+    *EXT_FILTER_LEVEL.write().unwrap() = level;
+}
+
+/// Sets log filter levels based on environment variables `RUST_LOG` and `CURSIVE_LOG`.
+/// If `RUST_LOG` is set, then both internal and external log levels are set to match.
+/// If `CURSIVE_LOG` is set, then the internal log level is set to match with precedence over
+/// `RUST_LOG`.
+pub fn set_filter_levels_with_env() {
```
gyscos, line 56: Nit
```suggestion
pub fn set_filter_levels_from_env() {
```

cursive-core/src/logger.rs
```diff
@@ -1,15 +1,71 @@
...
+    if let Ok(rust_log) = std::env::var("RUST_LOG") {
+        if let Ok(filter_level) = log::LevelFilter::from_str(&rust_log) {
```
gyscos, line 58: I'd log (ha!) a warning in case we can't parse the level, rather than silently ignoring it.

gyscos (commented):
Maybe we could use `VecDeque::with_capacity` from `lazy_static!` rather than manually calling `reserve_logs`. It'd still only be called on the first use (so no wasted memory if users don't use this logger).

Not sure we even need to expose `reserve_logs`? 🤷 

> What about the change to not having a set_log_size() function and just using reserve_logs()? (https://github.com/blaisdellma/cursive/pull/1#issuecomment-1461160889)

Yeah let's keep the API minimal, any power user should use a real serious solution.

cursive-core/src/logger.rs
```diff
@@ -1,15 +1,71 @@
...
+/// # use log::LevelFilter;
+/// logger::set_int_filter_level(LevelFilter::Warn);
+/// logger::set_ext_filter_level(LevelFilter::Debug);
+/// logger::init();
+/// ```
+
 pub struct CursiveLogger;
 
-static LOGGER: CursiveLogger = CursiveLogger;
+lazy_static! {
```
gyscos, line 35: Nit: let's combine this `lazy_static!` block and the one that defines `LOGS`?

cursive-core/src/logger.rs
blaisdellma: This is meant to be called before the logger is initialized, so any logged warning would be ignored. I guess you could call it after initialization, but then when initializing the `log::set_max_level()` would be based on the default filter levels (which is fine, cause it defaults to trace). I'll add it anyway.

blaisdellma:
> Not sure we even need to expose `reserve_logs`? shrug

I'm personally in favor of not making breaking changes to the API unless necessary. I don't know that it hurts to have it there, but it's your call.

gyscos:
Thanks again for all the work and updates, and sorry for the trouble!

PR #726, opened 2023-04-07
Fix examples for crossterm on windows

lesleyrs:
Hi, crossterm recently added release events for windows but that means that whenever checking for events you get double of them. This breaks almost all examples.

It's actually even worse because powershell has some kind of bug where it sends a delayed release event when pressing enter in the command line, and some examples immediately close because of this.

This fixes all of that by just ignoring all release events. https://github.com/crossterm-rs/crossterm/pull/745

Side note: please mention somewhere that you have to be inside the cursive/cursive folder in order to run examples with another backend, that confused me so much for a second time when using this because you get confusing errors about ncurses.

### DIFF
--- a/cursive/src/backends/crossterm.rs
+++ b/cursive/src/backends/crossterm.rs
@@ -19,7 +19,7 @@ use crossterm::{
     cursor,
     event::{
         poll, read, DisableMouseCapture, EnableMouseCapture, Event as CEvent, KeyCode,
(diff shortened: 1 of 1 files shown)


lesleyrs:
Closing this because it's being reverted https://github.com/crossterm-rs/crossterm/pull/778

lesleyrs:
@gyscos ignore the close, this is ready to be merged. It works even if the crossterm PR is merged.

gyscos:
Thanks for the work!

PR #739, opened 2023-07-06
Handle Ctrl-Key events in EditView & some general, small improvements

christoph-heiss:
Hi!

First of, thank you for `cursive`!
A rather important thing that popped up while implementing the new [TUI-based installer](https://git.proxmox.com/?p=pve-installer.git;a=tree;f=proxmox-tui-installer;hb=HEAD) for Proxmox was that the `EditView` does not handle `^U`, `^K` etc.

This implements that, plus some tests on top as well.
Also included is the addition of a `.get_cursor()` method, as well as the clamping of the length in `.remove()`, two things that would be quite useful to have.

Thanks in advance!

### DIFF
--- a/cursive-core/src/views/edit_view.rs
+++ b/cursive-core/src/views/edit_view.rs
@@ -381,6 +381,11 @@ impl EditView {
         self
     }
 
+    /// Returns the currest cursor position.
+    pub fn get_cursor(&self) -> usize {
+        self.cursor
+    }
+
     /// Sets the cursor position.
     pub fn set_cursor(&mut self, cursor: usize) {
         self.cursor = cursor;
--- a/cursive-core/src/views/edit_view.rs
+++ b/cursive-core/src/views/edit_view.rs
@@ -430,7 +430,7 @@ impl EditView {
     /// You should run this callback with a `&mut Cursive`.
     pub fn remove(&mut self, len: usize) -> Callback {
         let start = self.cursor;
-        let end = self.cursor + len;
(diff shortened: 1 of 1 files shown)


gyscos:
Hi, and thanks for the PR!

I wonder if these could be added with just an `OnEventView` wrapper around an `EditView`?
I'm not entirely sure these shortcuts are idiomatic for all users. So I'd love to make it easy for your app to use define them, but I don't think they belong in the default EditView configuration.

christoph-heiss:
Thanks for taking a look!

> I wonder if these could be added with just an `OnEventView` wrapper around an `EditView`?

Sure, that wouldn't be much trouble either.

> I'm not entirely sure these shortcuts are idiomatic for all users.

Well, there pretty much universal across TTY, CLIs and ncurses(-ish) applications/TUIs, so it is somewhat reasonable to expect them to work here too. Thus they are really not idiomatic w.r.t our usecases, that's why I chose to try to get them implemented upstream.

> So I'd love to make it easy for your app to use define them, but I don't think they belong in the default EditView configuration.

In a sense that `EditView` should be a rather "dumb" view, only implementing the bare necessities for text editing? I get that.

I'll drop the relevant commits if that's the way forward.

Are the other two commits (7538f11463a04940d928feb10e1077a7f9436770 and 120c19cb36b9e3fc631ed0408bc97c7673c30a37) acceptable? Especially the former, the latter would be simply nice to have.

gyscos:
Thanks for the work!

PR #741, opened 2023-07-13
(WIP) add wasm-backend using HTML canvas element

genieCS:


cursive/Cargo.toml
```diff
@@ -48,11 +48,33 @@ version = "2"
...
+optional = true
+version = "0.2"
+features = ["js"] 
+
 [features]
 doc-cfg = ["cursive_core/doc-cfg"]  # Enable doc_cfg, a nightly-only doc feature.
 builder = ["cursive_core/builder"]
 blt-backend = ["bear-lib-terminal"]
-default = ["ncurses-backend"]
+default = []
```
christoph-heiss, line 77: Any reason to remove the default backend? That just break things for people who don't select one explicitly? (And e.g. running examples from the repository)

cursive-core/Cargo.toml
```diff
@@ -64,13 +65,40 @@ default-features = false
...
 version = "0.9"
 
+[dependencies.js-sys]
+version = "0.3.64"
+optional = true
+
+[dependencies.getrandom]
+optional = true
+version = "0.2"
+features = ["js"]
```
christoph-heiss, line 75: Am I missing something or is this dependency never used anywhere? Possible development-leftover?

cursive/Cargo.toml
```diff
@@ -48,11 +48,33 @@ version = "2"
...
+    "TextMetrics",
+    "Window",
+    "Document",
+    "console",
+]
+
+[dependencies.getrandom]
+optional = true
+version = "0.2"
+features = ["js"]
```
christoph-heiss, line 71: Same question here as above.

cursive-core/Cargo.toml
```diff
@@ -64,13 +65,40 @@ default-features = false
...
+optional = true
+version = "0.2.87"
+
+[dependencies.wasm-bindgen-futures]
+optional = true
+version = "0.4.37"
+
 [features]
-default = []
+default = ["wasm"]
```
christoph-heiss, line 94: Also, this shouldn't be de default - again breaks the build due to selecting async-features (which are fundamentally incompatible with the other backends).

christoph-heiss (changes requested):
Just a shallow, quick review; currently breaks builds due to changing default features.

genieCS:
> Just a shallow, quick review; currently breaks builds due to changing default features.

thank you for taking a look:) I addressed what you mentioned

cursive/src/backends/wasm.rs
```diff
@@ -0,0 +1,266 @@
...
+                13 => cloned.borrow_mut().push_back(Event::Key(Key::Enter)),
+                37 => cloned.borrow_mut().push_back(Event::Key(Key::Left)),
+                38 => cloned.borrow_mut().push_back(Event::Key(Key::Up)),
+                39 => cloned.borrow_mut().push_back(Event::Key(Key::Right)),
+                40 => cloned.borrow_mut().push_back(Event::Key(Key::Down)),
+                code => {
+                    if let Some(c) = std::char::from_u32(code) {
+                        cloned.borrow_mut().push_back(Event::Char(c));
+                    }
+                }
```
christoph-heiss, line 98: What about all the other key-events, such as Escape, control/alt keys, etc? These are all also retrievable, going by the [`web_sys::KeyboardEvent`](https://docs.rs/web-sys/latest/web_sys/struct.KeyboardEvent.html) documentation.

Should be implemented too for completeness sake - as well as control-key-sequences are rather often used by TUI applications.

Also, rewriting/simplifying this to something like this form would be good readability-wise:
```suggestion
            let event = match event.key_code() {
                8 => Event::Key(Key::Backspace),
                ..
             };
             cloned.borrow_mut().push_back(event);
```

cursive/src/backends/wasm.rs
```diff
@@ -0,0 +1,266 @@
...
+        let color = (*self.color.borrow()).clone();
+        let mut buffer = self.buffer.borrow_mut();
+        for (i, c) in text.chars().enumerate() {
+            let x = pos.x + i;
+            buffer[self.width * pos.y + x] = TextColorPair::new(c, color.clone());
+        }
+    }
+
+    fn clear(self: &Backend, _color: cursive_core::theme::Color) {
+    }
```
christoph-heiss, line 186: Should be rather simple to implement too?
Just just be something like:

```suggestion
    fn clear(self: &Backend, color: cursive_core::theme::Color) {
        self.buffer.borrow_mut().fill(TextColorPair::new(' ', color.clone()));
    }
```
(untested tho)

christoph-heiss (changes requested):
Apart from two more comments, looks pretty good overall!

Can also confirm that builds and tests now run through again without a problem.

cursive/src/backends/wasm.rs
```diff
@@ -0,0 +1,304 @@
...
+    fn clear(self: &Backend, _color: cursive_core::theme::Color) {
+        let color = cursive_to_color(_color);
+        let pair = ColorPair {
+            front: color,
+            back: color,
+        };
+        let mut buffer = self.buffer.borrow_mut();
+        for i in 0..self.width * self.height {
+            buffer[i] = TextColorPair::new(' ', pair.clone());
+        }
```
christoph-heiss, line 223:
```suggestion
        self.buffer
            .borrow_mut()
            .fill(TextColorPair::new(' ', pair.clone()));
```
Using the provided `.fill()` is a better way, as it might get (easier) vectorized and generally benefits from optimizations in the standard library.

cursive/src/backends/wasm.rs
```diff
@@ -0,0 +1,304 @@
...
+        let color = (*self.color.borrow()).clone();
+        let mut buffer = self.buffer.borrow_mut();
+        for (i, c) in text.chars().enumerate() {
+            let x = pos.x + i;
+            buffer[self.width * pos.y + x] = TextColorPair::new(c, color.clone());
+        }
+    }
+
+    fn clear(self: &Backend, _color: cursive_core::theme::Color) {
+        let color = cursive_to_color(_color);
```
christoph-heiss, line 215:
```suggestion
    fn clear(self: &Backend, color: cursive_core::theme::Color) {
        let color = cursive_to_color(color);
```
`_color` doesn't need the `_` prefix anymore, is it's used now.

christoph-heiss (commented):
Really just some nits this time, but generally LGTM. :+1:

gyscos:
Thanks for the work!

My main concern is that the `async` features:
* is non-additive since it changes the API
* seems ties to wasm (we could imagine an async loop without wasm).

To avoid these two points, I'd be open to a major bump for API changes if needed, maybe even making async the default way to run the loop (or an option regardless of backend).

genieCS:
@gyscos I made run as async. Could you have another look?

gyscos:
(I'm finally back from a family trip and have at last some time to look at this PR in more depth.)

Thanks for the updates and reactivity!

Looking at the code more, I _think_ we can actually avoid breaking changes by adding a parallel code-path for async (rather than replacing the existing one). The current "sync" API could be defined as simply calling `block_on` the async API, preserving compatibility with existing code, without duplicating (too much) logic (I wish this didn't require the `futures` crate though).

The "bare minimum" would be to have `async fn post_events_async()`, so users could re-build their own event loop (the various `step` and `run` methods are just convenient methods to remove some boilerplace, but could be re-implemented user-side).

We could then, later, add the convenient equivalent.

It also looks like the `sleep` behaviour should maybe be part of the backend instead - and this opens up the possibility of having other async methods from the backend (input? drawing? refresh?). At this point, `post_events` will no longer be the only method with an async equivalent.
The main issue is having async in traits currently requires the `async_trait` crate. The heap allocation per method from async-trait doesn't matter much when it's just to call `run` (or even `step`), but it might not be ideal for the Backend trait, especially for the many `draw` methods (potentially thousands per second). Cannot wait for `async_fn_in_trait` (rust 1.74? right around the corner!).

As a side-question, I'm wondering if this wasm backend would work in a non-browser wasm environment (like wasmtime?). I suspect not, so I'd favor calling it `wasm_browser` or `wasm_js` or something else more indicative of the browser aspect.

(I'm really sorry for the back-and-forth.)

genieCS:
@gyscos thank you for your time and effort to review this PR.

* "wasm_js" or "wasm_browser" idea sounds good to me. I'll think more to rename it.

* Promise can be executed only if it runs forefront.
In your `async-restricted` branch, it is wrapped by `futures::executor::block_on` so it would not work.
How about having duplicate sets of `sleep`, `post_events`, `run`, `step`, ...etc with configuration option as my first draft?

regarding test, I'll make a simple project and let you know in a week

genieCS:
@gyscos I made a simple project(https://github.com/genieCS/hello-wasm) for tests.
1. in `/` path, run `wasm-pack build`
2. in `www/`path, run `npm run start`
if you see "Hello from Rust!" in web console, it works.

genieCS:
@gyscos  canvas is for browser but it would not be hard to make other wasm backends work.
I'd like to work on implementing other backend as a separate PR.

regarding `async_trait`, I agree that it would be more proper to put sleep in backend. 
but for now, it's only `sleep`, so how about keep current status and when we have more to change Backend trait related to async, change together?

PR #743, opened 2023-07-18
Add missing Shift-Tab key event in Termion backend

christoph-heiss:
This fixes Shift-Tab detection (and thus circular focus view etc.) when using the Termion backend.

Greets!


gyscos:
Thanks for the work!

PR #744, opened 2023-07-28
Add method to return character from char event.

timdubbins:
Firstly, thank you for this library. I've really enjoyed hacking my way through it.

This is a minor proposal that helped me clean up some ugly code. Here's an example:

``` Rust
use cursive::event::{Event, EventResult, EventTrigger};
use cursive::views::TextView;

fn main() {
    let mut siv = cursive::default();
    siv.add_layer(TextView::new("enter uppercased character:"));

    siv.set_on_pre_event_inner(uppercased(), move |event| {
        // use the method here
        let c = event.char().expect("only chars are triggered");
        Some(EventResult::with_cb(move |siv| {
            siv.add_layer(TextView::new(format!("do something useful with: {}", c)))
        }))
    });

    siv.set_on_pre_event(Event::Char('q'), |siv| siv.quit());
    siv.run();
}

fn uppercased() -> EventTrigger {
    EventTrigger::from_fn(|e| matches!(e, Event::Char('A'..='Z')))
}
```

This `Event::chars()` method allows us to access the inputted character without creating a mapping, which is a bit tedious when you want to capture a range of values (as we do in the example).

### DIFF
--- a/cursive-core/src/event.rs
+++ b/cursive-core/src/event.rs
@@ -533,6 +533,15 @@ pub enum Event {
 }
 
 impl Event {
+    /// Returns the character, if `self` is a char event.
+    pub fn char(&self) -> Option<char> {
+        if let Event::Char(c) = *self {
+            Some(c)
+        } else {
+            None
+        }
+    }
+    
     /// Returns the position of the mouse, if `self` is a mouse event.
     pub fn mouse_position(&self) -> Option<Vec2> {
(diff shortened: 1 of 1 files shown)


gyscos:
Thanks for the PR!
I think in this case a `match self` might be shorter and simpler than a if let.

Also, should it maybe also return the `char` for AltChar and CtrlChar events?

gyscos (approved):

PR #750, opened 2023-08-27
crossterm: remove unused import when not using "windows"

hasezoey:
This PR removes a unused import when not using target `windows` for `std::io::{self}` for module `crossterm`


gyscos:
Thanks for the PR!

gyscos (approved):

PR #761, opened 2023-09-25
Scroll the whole page on PageDown/PageUp

azat:


gyscos:
Hi, and thanks for the PR!

I'd probably scroll by the screen height _minus 1_ (but with a minimum of 1), so the last line before pressing scroll-down is still visible at the top. What do you think?

azat:
> I'd probably scroll by the screen height minus 1 (but with a minimum of 1), so the last line before pressing scroll-down is still visible at the top. What do you think?

Initially I did the same as `less` does - it scroll the whole page page (on PageUp/PageDown).
I think that here it is better to preserve behavior of some existing tool, and `less` looks like a good example to follow.

But if you still think that the last line should be preserved, I've pushed a patch on top that does this, feel free to apply either only the first one, or the whole patch set.

gyscos:
Indeed, it seems different application use different configurations:
* Pagers, such as `less`, `more`, `bat` seem to keep no line, and show an entire "new page" with page-down (or space).
* Editors seem to preserve some lines (my `nvim` seems to preserve 2 lines, my gedit 5 lines, ...).

I agree that scroller should probably behave more like a pager than an editor (editable views like `TextArea` would implement their own page-down handling).

gyscos:
I merged the first commit.

PR #764, opened 2023-10-20
Scroll the whole page on Left/Right arrows

azat:
Seeking char by char can take awhile... Or maybe it is done on purpose?

### DIFF
--- a/cursive-core/src/view/scroll/raw.rs
+++ b/cursive-core/src/view/scroll/raw.rs
@@ -313,12 +313,14 @@ pub fn on_event<Model: ?Sized>(
                 Event::Ctrl(Key::Left) | Event::Key(Key::Left)
                     if get_scroller(model).can_scroll_left() =>
                 {
-                    get_scroller(model).scroll_left(1);
+                    let scroller = get_scroller(model);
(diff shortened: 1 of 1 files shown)


michalfita:
@azat this looks like semantic change to the existing behaviour; it's a candidate for different shortcut.

TBH: I'm surprised Ctrl+Arrow does just scrolling by the same amount than normal Arrow, I'd expect Ctrl and Shift to modify amount of the movement.

azat:
> this looks like semantic change to the existing behaviour; it's a candidate for different shortcut.

Scrolling just one column sounds odd anyway, for instance `less` does not do this (but it seeks half of screen by default, not the whole screen, while it does scroll the whole screen for vertical scroll).

> TBH: I'm surprised Ctrl+Arrow does just scrolling by the same amount than normal Arrow, I'd expect Ctrl and Shift to modify amount of the movement.

So as I, I was thinking about using Ctrl+Left/Right instead, but decided that the default looks odd anyway and it could be adjusted, but I'm OK with using different binding for this as well.

azat:
Kind ping on this one

PR #765, opened 2023-10-25
[Fixes #763] Introduced labelled checkboxes

### LINKED ISSUE #763: [FEATURE] `CheckBox` as multichoice with similarity to `RadioGroup`
**Is your feature request related to a problem? Please describe.**
Looks like the `CheckBox` is not a popular widget. I found it bizarre that contrary to radio box it's not linked with it's label (I have to make `LinearLayout` with `TextView` on the right, but then highlighting for selection work only over check boxes, not the whole line.

On top of that would be nice to have some kind of `MultichoiceGroup` control, returning `HashSet<T>` or `HashSet<Rc<T>>` with selections.

**Describe the solution you'd like**
* Ability to visually link a checkbox with it's label for nicer highlight on focus.
* Grouping feature to manage multi-choice as single widget with result (higher level of abstraction).
* Ability to skip disabled items when cursor is moved like `RadioGroup` does.

**Describe alternatives you've considered**
I brutally encoded bunch of `CheckBox`es with `TextView`s linked by horizontal `LinearLayout`, but it's not visually appealing on focus. Then I'd have to handle evens individually to maintain a state of possible selections.

**Additional context**
image

I'm willing to provide PR if API of `CheckBox::new()` can be aligned with `RadioBox::new()`, or a new _builder_ `::labelled()` can be introduced and we let both widgets be created w/o attached label, so `RadioBox::new()` looses the argument and `label` member becomes `Optional<StyledString>`.

michalfita:
## Purpose
To address #763 and improve overall UX of using `Checkbox` in the application.

## Scope
* The implementation
* Examples

## To do
- [x] Add label for showing selection over whole `Checkbox`
- [ ] Create grouping for `MultiChoice` to make application logic easier
- [ ] Example combining `Checkbox` and `RadioButton`

## Notes
Commits bring changes incrementally, so I suggest reviewing them one by one and consider rejections based on changes in them as maybe not all PR would need to be rejected at once.

For now I do _draft_ for first impression where I'm heading. I'm trying to not break original API, however, some alignment with `RadioButton` in future would do good.

### DIFF
--- a/cursive-core/src/views/checkbox.rs
+++ b/cursive-core/src/views/checkbox.rs
@@ -3,7 +3,7 @@ use crate::{
     event::{Event, EventResult, Key, MouseButton, MouseEvent},
     theme::PaletteStyle,
     view::{CannotFocus, View},
-    Cursive, Printer, Vec2, With,
+    Cursive, Printer, Vec2, With, utils::markup::StyledString,
 };
 use std::rc::Rc;
 
@@ -25,6 +25,8 @@ pub struct Checkbox {
     enabled: bool,
 
     on_change: Option<Rc<Callback>>,
+
+    label: StyledString,
 }
 
 new_default!(Checkbox);
@@ -32,12 +34,23 @@ new_default!(Checkbox);
 impl Checkbox {
     impl_enabled!(self.enabled);
 
-    /// Creates a new, unchecked checkbox.
+    /// Creates a new, unlabelled, unchecked checkbox.
     pub fn new() -> Self {
         Checkbox {
             checked: false,
             enabled: true,
             on_change: None,
+            label: StyledString::new(),
+        }
+    }
+
+    /// Creates a new, labelled, unchecked checkbox.
+    pub fn labelled(label: StyledString) -> Self {
+        Checkbox {
+            checked: false,
+            enabled: true,
+            on_change: None,
+            label
         }
     }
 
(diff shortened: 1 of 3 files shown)


gyscos:
Thanks for the work!

I merged main into this branch (solving minor conflicts that arose when views became `Send`), and it should now be closer to ready.

One remark: the example now shows 2 ways to make groups of checkboxes: the "manual way" (for toppings) and the "managed way" (for extras). To keep the example simple, I think we can get rid of the manual way and only show how to use the multichoice group.

Instead, we could include a couple of "non-grouped" checkboxes, to show that it's still an option for single options. Or not, since that's already covered in the "list_view" example. In that case, I would rename the checkbox example to checkbox_group or multichoice_group.

michalfita:
@gyscos Oh man, it's been **a while** since I've been on this... Finding time to go back to this may be close to impossible in a next few weeks.

cursive-core/src/views/checkbox.rs
```diff
@@ -1,14 +1,122 @@
...
+            move |_, checked| {
+                if checked {
+                    if let Some(v) = selectable.upgrade() {
+                        groupstate.lock().selections.insert(v);
+                    }
+                } else if let Some(v) = selectable.upgrade() {
+                    groupstate.lock().selections.remove(&v);
+                }
+            }
+        })
```
gyscos: If you don't need to keep the Arc around, it gets a bit simpler:
```suggestion
        let i = self.state.lock().add(value);
        
        Checkbox::labelled(label).on_change({
            let groupstate = Arc::clone(&self.state);
            move |_, checked| {
                groupstate.lock().set_checked(i, checked);
            }
        })
```

cursive-core/src/views/checkbox.rs
```diff
@@ -140,12 +261,26 @@ impl Checkbox {
...
+            printer.print_styled((4, 0), &self.label);
+        }
+    }
+
+    fn req_size(&self) -> Vec2 {
+        if self.label.is_empty() {
+            Vec2::new(3, 1)
+        } else {
+            Vec2::new(3 + 1 + self.label.width(), 1)
+        }
```
gyscos: Why a separate function? Let's just move the implementation to `required_size`?

gyscos (commented):
Thanks again for the PR!

Just a few minor details.

Also, are the two examples different?

michalfita:
> Also, are the two examples different?

Now, examples are even more different. I think they'd work well for someone trying to consider various options how to use checkboxes.

gyscos:
Thanks again for the work!

PR #768, opened 2023-12-04
Add ability to replace an item in a select view

squaresurf:
This is a more efficient way to replace an item than calling `remove_item`, then `insert_item`

### DIFF
--- a/cursive-core/src/views/select_view.rs
+++ b/cursive-core/src/views/select_view.rs
@@ -424,6 +424,15 @@ impl<T: 'static> SelectView<T> {
             .unwrap_or_else(Callback::dummy)
     }
 
+    /// Replaces an item in the list.
+    pub fn replace_item<S>(&mut self, id: usize, label: S, value: T)
+    where
+        S: Into<StyledString>,
+    {
+        self.items[id] = Item::new(label.into(), value);
+        self.last_required_size = None;
+    }
+
     /// Inserts an item at position `index`, shifting all elements after it to
     /// the right.
     pub fn insert_item<S>(&mut self, index: usize, label: S, value: T)


gyscos:

Could [`SelectView::get_item_mut()`](https://docs.rs/cursive/latest/cursive/views/struct.SelectView.html#method.get_item_mut) solve your use-case? Or [`SelectView::iter_mut`](https://docs.rs/cursive/latest/cursive/views/struct.SelectView.html#method.iter_mut) if you need to access several items. In particular, you should be able to swap two items efficiently this way.

I'm fine adding this method, if only for the `S: Into<StyledString>` convenience factor. Wondering if it should be `set_item` vs `replace_item`, also, we could return the old value? :shrug:

squaresurf:
> Could [SelectView::get_item_mut()](https://docs.rs/cursive/latest/cursive/views/struct.SelectView.html#method.get_item_mut) solve your use-case? Or [SelectView::iter_mut](https://docs.rs/cursive/latest/cursive/views/struct.SelectView.html#method.iter_mut) if you need to access several items. In particular, you should be able to swap two items efficiently this way.

I'm not sure why, but I couldn't seem to get a mutable reference properly where the compiler allowed me to mutate the references. Then I realized that I didn't care about the old values and just need an efficient way to replace an item.

> Wondering if it should be set_item vs replace_item
Great idea. I'm happy to make this change.

> also, we could return the old value? 🤷
I like this as it feels consistent with other set_item type APIs. I gave it a shot, but it turns out to be a bit complicated to return the values within the `Rc`. I've still not really used `Rc` much and am running into many issues with working with the type and the compiler. If you have a chance, would you mind adding a suggestion in the diff so I could see how to do this?

squaresurf:
> I'm fine adding this method, if only for the S: Into<StyledString> convenience factor.

Do you mind explaining what you mean by this convenience? I'm not sure I understand this.

squaresurf:
@gyscos I did not disagree with your suggestions in your last comment, but have not had time to address it. I am also moving away from my usage of cursive so I won't have the time in the future. Thank you so much for creating cursive as it was a ton of fun to work with. Feel free to pick up this PR if you find it useful.

p.s. my move away from cursive is more so a movement from away from building my TUI app and not a critique of cursive.

PR #771, opened 2024-01-16
fixed links for examples in cursive/Readme.md

Jadens-arc:
Links for example programs were not linking to example code correctly. Now fixed


gyscos:
I ended up using absolute links, as it might be the only way to work on both github's front page and the crates-based pages.

PR #778, opened 2024-04-11
make `cursive` work with `ncurses` v6.0.1

correabuscar:
<del>but first: `ncurses-rs` needs the changes in this PR: https://github.com/jeaye/ncurses-rs/pull/220  (or ideally the ones in this PR https://github.com/jeaye/ncurses-rs/pull/218 , depending on which PR gets merged I'll (re)test the below)</del>

`ncurses` crate v 6.0.1 is published and enough.

However `pancurses` doesn't need [these](https://github.com/ihalila/pancurses/pull/93) changes looks like, for `cursive` to work with it, so `0.17` seems good (but `cursive`'s examples are not tested with `0.17`).

The following were used successfully:
* `cargo build`
* `cargo test`
* `cargo build --all-features`
* `cargo test --all-features`
* `cargo build --all-targets`
* `cargo test --all-targets`
* `cargo build --all-targets --all-features`
* `cargo test --all-targets --all-features`
* `cargo build --features="ncurses-backend"`
* `cargo test --features="ncurses-backend"`
* `cargo build --features="ncurses-backend" --all-targets`
* `cargo test --features="ncurses-backend" --all-targets`

... on each of these target environments:

Legend:
- ❌ = known not to work
- [ ] = untested
- [x] = works

- [x] Gentoo 2.15 default/linux/amd64/23.0/split-usr/no-multilib (stable) <del>-won't build but it would've if https://github.com/jeaye/ncurses-rs/pull/218 was merged.</del>
  - [ ] some examples tested
  - [ ] all examples tested
- [ ] NixOS (if `PKG_CONFIG_PATH` is set to the dir containing `ncurses.pc` file, or used `nix-shell` command in `cursive` repo dir)
  - [ ] some examples tested
  - [ ] all examples tested
- [ ] Fedora 39
  - [ ] some examples tested
  - [ ] all examples tested
- [ ] Ubuntu Desktop 22.04.4 LTS (needs `libncurses-dev`)
  - [ ] tested to work without `pkg-config` and `pkgconf` installed.
  - [ ] tested to work with `pkg-config` installed.
- [ ] FreeBSD 14.0-RELEASE
  - [ ] some examples tested
  - [ ] all examples tested
- [ ] OpenBSD 7.5 GENERIC
  - [ ] needs one of `LC_CTYPE=en_US.UTF-8` or `LANG=en_US.UTF-8` to be set otherwise it looks as if it doesn't have wide chars support so `cursive` and `pancuses` examples look pretty broken.
- [ ] MacOS Mojave v10.14.6 (with [this](https://github.com/gyscos/cursive/wiki/Install-ncurses#macos))
  - [ ] some examples tested
  - [ ] all examples tested
- [ ] Windows 11 WSL1 [Arch Linux](https://github.com/yuk7/ArchWSL) Linux DESKTOP 4.4.0-22621-Microsoft #2506-Microsoft Fri Jan 01 08:00:00 PST 2016 x86_64 GNU/Linux
  - [ ] some examples tested
  - [ ] all examples tested

TODO:
- [x] compiles(and passes `cargo test`) with `pancurses` (dep) `0.17` but this means it uses `v5.101.0` of `ncurses` crate for `pancurses` and v 6.0.1 for `cursive`, see `cargo tree --all-features | less` (search for `ncurses`)
- [x] compiles(and passes `cargo test`) with `pancurses` `main` branch from github with [this](https://github.com/ihalila/pancurses/pull/93) PR in.
- [ ] does `cursive` need version bump in this PR? ie. `version = "0.21.0"` or shall this be left to be done by the repo owner ?!
- [x] use exact `ncurses-rs` version in `Cargo.toml` after it gets published, ie. so it's not lower than that version, because then it would break compilation of `cursive`. Note that `pancurses` doesn't need a version bump (see above), although the examples I'm running are through the PR-ed `pancurses` only.
- [ ] maybe bump `pancurses` in `Cargo.toml` too, after it gets [its PR](https://github.com/ihalila/pancurses/pull/93) for the v6 `ncurses-rs` usage.
- [x] add `cargo test --example select_test -- --nocapture` in addition to the suggested `--bin` which doesn't work.
- [ ] fix all warnings (because they were introduced by this PR / the upgrade to `ncurses-rs` v6)
  - [ ] `warning: attribute should be applied to a foreign function or static` `#[link_name="box"]` `not a foreign function or static`, [this is on `ncurses-rs`](https://github.com/jeaye/ncurses-rs/pull/218#issuecomment-2049572573)
  - [x] `warning: unused `Result` that must be used`, (for all 3 cases) happens due to [`CString::new`](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.new) `This function will return an error if the supplied bytes contain an internal 0 byte.`
    - [x] `ncurses::newterm` `.unwrap()` is safe here due to using `None` as the first arg to [`newterm`](https://github.com/jeaye/ncurses-rs/blob/3aa22bc279e4929e3ab69d49f75a18eda3e431e9/src/lib.rs#L1023-L1029)
    - [x] `ncurses::addstr`
      - [x] `\0` is already handled by `cursive` (due to PR #786 and followups) so it won't reach the `addstr` call, <del>unless in the future a dep gets updated to include this https://github.com/unicode-rs/unicode-width/commit/4efb1803faa054f1bea3c0457275ad3c8610170b#diff-2ad10836ccce5ac2056d5679cc92449d9ff9094d4ff5c5803f65b5dd1d52ef19R224 </del> this is handled too now.
- [ ] run `cargo clippy` and fix:
  - [x] errors (there were none)
  - [ ] warnings
    - [ ] pre this PR warnings
      - [ ] via `cargo clippy --all-targets`
      - [ ] via `cargo clippy --all-targets --all-features`
- [x] passes `cargo test --features="ncurses-backend,ansi,toml"`
- [x] get rid of old irrelevant commits that this PR introduced, since they're now mostly obsolete due to new `cursive` commits.
- [x] I'm watching(github notifications All Activity) the entire repo for any future-reported breakage, to address it.
- [ ] test that examples compile and run seemingly ok(for  on which OS/distro see above) via `cargo run --features="ncurses-backend" --example NAME`:
  - [x] advanced_user_data
  - [x] ansi via `cargo run --features="ncurses-backend,ansi" --example ansi`
  - [x] autocomplete
  - [x] builder via `cargo run --features="ncurses-backend,builder" --example builder` (but looks better with `crossterm-backend` which is now the default I [noticed](https://github.com/gyscos/cursive/commit/7c676b29f47b6c4e79f8d67ef5509c047f08edd4))
  - [x] colors
  - [x] ctrl_c
  - [x] debug_console
  - [x] dialog
  - [x] edit
  - [x] fixed_layout
  - [x] focus
  - [x] hello_world
  - [x] key_codes
  - [x] linear
  - [x] list_view
  - [x] logs (note: the default aka `crossterm-backend` has no flickering on bottom line which is great)
  - [x] lorem
  - [x] markup
  - [x] menubar
  - [ ] mines
  - [ ] mutation
  - [ ] panels
  - [x] parse
  - [ ] pause
  - [ ] position
  - [ ] progress
  - [ ] radio
  - [ ] refcell_view
  - [ ] scroll
  - [ ] select
  - [x] `select_test` via `cargo test --features="ncurses-backend" --example select_test -- --nocapture`
  - [ ] slider
  - [ ] status
  - [ ] status_bar_ext
  - [ ] stopwatch
  - [ ] tcp_server
  - [ ] terminal_default
  - [ ] text_area
  - [ ] theme
  - [ ] theme_editor
  - [ ] theme_manual
  - [x] themed_view
  - [x] user_data
  - [x] vpv
  - [x] window_title
- [ ] test each of those in [showcases](https://github.com/gyscos/cursive?tab=readme-ov-file#showcases):
  - [ ] [RustyChat](https://github.com/SambaDialloB/RustyChat): Chat client made using Rust and Cursive.
  - [ ] [checkline](https://github.com/sixarm/checkline-rust-crate): Checkbox line picker from stdin to stdout.
  - [ ] [clock-cli](https://github.com/TianyiShi2001/clock-cli-rs): A clock with stopwatch and countdown timer functionalities.
  - [ ] [fui](https://github.com/xliiv/fui): Add CLI & form interface to your program.
  - [ ] [game2048-rs](https://github.com/genieCS/game2048-rs): a tui game2048 using Rust and cursive.
  - [ ] [git-branchless](https://github.com/arxanas/git-branchless): Branchless workflow for Git.
  - [ ] [grin-tui](https://github.com/mimblewimble/grin): Minimal implementation of the MimbleWimble protocol.
  - [ ] [kakikun](https://github.com/file-acomplaint/kakikun): A paint and ASCII art application for the terminal.
  - [ ] [launchk](https://github.com/mach-kernel/launchk): Manage launchd agents and daemons on macOS.
  - [ ] [mythra](https://github.com/deven96/mythra): CLI to search for music.
  - [ ] [ncspot](https://github.com/hrkfdn/ncspot): Cross-platform ncurses Spotify client.
  - [ ] [rbmenu-tui](https://github.com/DevHyperCoder/rbmenu-tui): A TUI for bookmark management.
  - [ ] [retris](https://github.com/genieCS/retris): A simple implementation of the classic tetris game.
  - [ ] [ripasso](https://github.com/cortex/ripasso): A simple password manager written in Rust.
  - [ ] [rusty-man](https://sr.ht/~ireas/rusty-man): Browse rustdoc documentation.
  - [ ] [saci-rs](https://gitlab.com/ihercowitz/saci-rs): Simple API Client Interface.
  - [ ] [so](https://github.com/samtay/so): A terminal interface for Stack Overflow.
  - [ ] [sudoku-tui](https://github.com/TianyiShi2001/sudoku-tui): Play sudoku on the command line.
  - [ ] [tap](https://github.com/timdubbins/tap): An audio player for the terminal with fuzzy finder.
  - [ ] [ttyloop](https://github.com/gamma-delta/ttyloop): Clone of the mobile game Loop.
  - [ ] [wiki-tui](https://github.com/Builditluc/wiki-tui): A simple and easy to use Wikipedia Text User Interface

### DIFF
--- a/cursive/src/backends/curses/n.rs
+++ b/cursive/src/backends/curses/n.rs
@@ -357,28 +357,28 @@ impl backend::Backend for Backend {
 
     fn set_effect(&self, effect: Effect) {
         let style = match effect {
-            Effect::Reverse => ncurses::A_REVERSE(),
-            Effect::Simple => ncurses::A_NORMAL(),
-            Effect::Dim => ncurses::A_DIM(),
-            Effect::Bold => ncurses::A_BOLD(),
-            Effect::Blink => ncurses::A_BLINK(),
-            Effect::Italic => ncurses::A_ITALIC(),
-            Effect::Strikethrough => ncurses::A_NORMAL(),
-            Effect::Underline => ncurses::A_UNDERLINE(),
+            Effect::Reverse => ncurses::A_REVERSE,
+            Effect::Simple => ncurses::A_NORMAL,
+            Effect::Dim => ncurses::A_DIM,
+            Effect::Bold => ncurses::A_BOLD,
+            Effect::Blink => ncurses::A_BLINK,
+            Effect::Italic => ncurses::A_ITALIC,
+            Effect::Strikethrough => ncurses::A_NORMAL,
+            Effect::Underline => ncurses::A_UNDERLINE,
         };
         ncurses::attron(style);
     }
 
     fn unset_effect(&self, effect: Effect) {
         let style = match effect {
-            Effect::Reverse => ncurses::A_REVERSE(),
-            Effect::Simple => ncurses::A_NORMAL(),
-            Effect::Dim => ncurses::A_DIM(),
-            Effect::Bold => ncurses::A_BOLD(),
-            Effect::Blink => ncurses::A_BLINK(),
-            Effect::Italic => ncurses::A_ITALIC(),
-            Effect::Strikethrough => ncurses::A_NORMAL(),
-            Effect::Underline => ncurses::A_UNDERLINE(),
+            Effect::Reverse => ncurses::A_REVERSE,
+            Effect::Simple => ncurses::A_NORMAL,
+            Effect::Dim => ncurses::A_DIM,
+            Effect::Bold => ncurses::A_BOLD,
+            Effect::Blink => ncurses::A_BLINK,
+            Effect::Italic => ncurses::A_ITALIC,
+            Effect::Strikethrough => ncurses::A_NORMAL,
+            Effect::Underline => ncurses::A_UNDERLINE,
         };
         ncurses::attroff(style);
     }
@@ -533,7 +533,7 @@ fn initialize_keymap() -> HashMap<i32, Event> {
     }
 
     // Ncurses provides a F1 variable, but no modifiers
-    add_fn(ncurses::KEY_F1, Event::Key, &mut map);
+    add_fn(ncurses::KEY_F(1), Event::Key, &mut map);
     add_fn(277, Event::Shift, &mut map);
     add_fn(289, Event::Ctrl, &mut map);
     add_fn(301, Event::CtrlShift, &mut map);
--- a/cursive/src/backends/curses/n.rs
+++ b/cursive/src/backends/curses/n.rs
@@ -110,7 +110,10 @@ impl Backend {
             let path = CString::new(output_path).unwrap();
             unsafe { libc::fopen(path.as_ptr(), mode.as_ptr()) }
         };
-        ncurses::newterm(None, output, input);
+        ncurses::newterm(None, output, input).map_err(|e| {
+            io::Error::new(io::ErrorKind::Other, format!("could not call newterm: {e}"))
+        })?;
+
         // Enable keypad (like arrows)
         ncurses::keypad(ncurses::stdscr(), true);
 
--- a/cursive/src/backends/curses/n.rs
+++ b/cursive/src/backends/curses/n.rs
@@ -405,7 +405,9 @@ impl backend::Backend for Backend {
     }
 
     fn print(&self, text: &str) {
-        ncurses::addstr(text);
+        // &str is assured it doesn't contain any \0 aka nuls here due to PR 786
+        // thus we can ignore the return value and avoid warning: unused `Result` that must be used
+        let _ = ncurses::addstr(text);
     }
 }
 
--- a/cursive/Cargo.toml
+++ b/cursive/Cargo.toml
@@ -34,7 +34,7 @@ version = "2"
 [dependencies.ncurses]
 features = ["wide"]
 optional = true
-version = "5.99.0"
+version = "6.0.1"
 
 [dependencies.pancurses]
 features = ["wide"]


correabuscar:
Question: how would you like to handle the case when that `text` &str there as the arg to `print_at` and `print_at_rep`, contains a `\0` (null byte) as part of the string, which is when [`CString::new`](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.new) from `ncurses-rs` side will return early with an `Err` for each of those 3 function calls in the code section below: [`ncurses::mvaddstr`](https://github.com/jeaye/ncurses-rs/blob/3aa22bc279e4929e3ab69d49f75a18eda3e431e9/src/lib.rs#L720-L724), [`ncurses::addstr`](https://github.com/jeaye/ncurses-rs/blob/3aa22bc279e4929e3ab69d49f75a18eda3e431e9/src/lib.rs#L98-L99):

https://github.com/gyscos/cursive/blob/0a95c82c88c8a791d7fd983e7fb9f568b77551a8/cursive/src/backends/curses/n.rs#L396-L405

I guess the options are:
- [ ] ignore them as:
  - [ ] leave them as they are as warnings
  - [ ] fix the warnings so prefix with `let _ = `
- [ ] panic via calling `.unwrap()` on the functions
- [ ] change `print_at*` signatures to gracefully pass the error to caller
  - [ ] presumably this? `-> Result<(), std::ffi::NulError>`
  - [ ] not this? `-> Result<(i32), std::ffi::NulError>`
- [ ] replace all `\0` from the `text` &str, how? by creating the resulting String on heap?
- [ ] something else ?

cursive/src/backends/curses/n.rs
```diff
@@ -396,19 +396,30 @@ impl backend::Backend for Backend {
     }
 
     fn print_at(&self, pos: Vec2, text: &str) {
-        ncurses::mvaddstr(pos.y as i32, pos.x as i32, text);
+        let text=&text.replace('\0'," "); // or the unwrap() will panic
```
correabuscar, line 399: In retrospect, replacing `\0` with spaces like this might not be a good idea due to this causing heap allocation which may not be desired considering the arg is `&str`. Maybe finding another way here(and below) would be in order?

correabuscar: actually nevermind, because internally [`mvaddstr`](https://github.com/jeaye/ncurses-rs/blob/3aa22bc279e4929e3ab69d49f75a18eda3e431e9/src/lib.rs#L720-L725) does eventually call [`CString::new`](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.new) [here](https://github.com/jeaye/ncurses-rs/blob/3aa22bc279e4929e3ab69d49f75a18eda3e431e9/src/lib.rs#L65) which does allocate memory on the heap. So, it still ends up allocating heap anyway.

gyscos:
Hi, and thanks a lot for the work! Sorry for the delay.

My main concern is, as you mentioned, the unconditional re-allocation. The fact that `ncurses` re-allocates as well is not an excuse, because:
* It's still an _extra_ re-allocation, making something non-ideal even worse.
* We could (should?) one day switch to using `addnstr` which might not need to be zero-terminated (though the current rest wrapper still re-allocates it).
* Other backends might not re-allocate.

Using a smarter "replace if" method that returns something like a `Cow` could be an option to avoid allocation in the general case, only paying the price when there's actually a `\0`.

Another option would be to only keep the string prefix up to that `\0`, ignoring the rest (this is what would have happened using the C library directly).

correabuscar:
Seems like, there's no need to use `.unwrap()` in some places(in this PR) just to get rid of the unused warnings, especially since the results were ignored before also, because counting of `.unwrap()` to not panic, depends on the implementation (of `ncurses-rs`) at the time, but if this changes in the future and it can return new errors for other reasons, then `.unwrap()` will panic.

I was too noob to realize that a `let _ =` would be enough.

Making the changes...

cursive/src/backends/curses/n.rs
```diff
@@ -393,22 +399,53 @@ impl backend::Backend for Backend {
...
             while dupes_left > 0 {
-                ncurses::addstr(text);
+                let _ = ncurses::addnstr(text, len);
                 dupes_left -= 1;
             }
         }
     }
 }
 
+#[inline(always)]
```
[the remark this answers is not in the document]
correabuscar: here do you mean remove the `always` or remove the whole line? I guess just the `always`...

cursive/src/backends/curses/n.rs
```diff
@@ -393,22 +399,53 @@ impl backend::Backend for Backend {
...
+        // https://github.com/gyscos/cursive/pull/778#discussion_r1613859129
+        let text = &delete_nuls(text);
+        let len = text.len() as i32;
+        // Ignore the value to avoid warning: unused `Result` that must be used
+        let _ = ncurses::mvaddnstr(pos.y as i32, pos.x as i32, text.as_ref(), len);
     }
 
     fn print_at_rep(&self, pos: Vec2, repetitions: usize, text: &str) {
+        // Remove '\0' from &str or else nothing would get printed
+        let text = &delete_nuls(text);
```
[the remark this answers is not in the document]
correabuscar: I knew I used `&` for a reason there:
```rust
   Compiling cursive v0.20.0 (/home/user/1tmp/ncurses_things/cursive/cursive)
error[E0308]: mismatched types
   --> /home/user/1tmp/ncurses_things/cursive/cursive/src/backends/curses/n.rs:417:68
    |
417 |             let _ = ncurses::mvaddnstr(pos.y as i32, pos.x as i32, text, len);
    |                     ------------------                             ^^^^ expected `&str`, found `Cow<'_, str>`
    |                     |
    |                     arguments to this function are incorrect
    |
    = note: expected reference `&str`
                    found enum `Cow<'_, str>`
note: function defined here
   --> /home/user/1tmp/ncurses_things/ncurses-rs/src/lib.rs:718:8
    |
718 | pub fn mvaddnstr(y: i32, x: i32, s: &str, n: i32) -> Result<i32, std::ffi::NulError>
    |        ^^^^^^^^^
help: consider borrowing here
    |
417 |             let _ = ncurses::mvaddnstr(pos.y as i32, pos.x as i32, &text, len);
    |                                                                    +

error[E0308]: mismatched types
   --> /home/user/1tmp/ncurses_things/cursive/cursive/src/backends/curses/n.rs:420:42
    |
420 |                 let _ = ncurses::addnstr(text, len);
    |                         ---------------- ^^^^ expected `&str`, found `Cow<'_, str>`
    |                         |
    |                         arguments to this function are incorrect
    |
    = note: expected reference `&str`
                    found enum `Cow<'_, str>`
note: function defined here
   --> /home/user/1tmp/ncurses_things/ncurses-rs/src/lib.rs:94:8
    |
94  | pub fn addnstr(s: &str, n: i32) -> Result<i32, std::ffi::NulError>
    |        ^^^^^^^
help: consider borrowing here
    |
420 |                 let _ = ncurses::addnstr(&text, len);
    |                                          +
```

cursive/src/backends/curses/n.rs
```diff
@@ -110,7 +111,12 @@ impl Backend {
             let path = CString::new(output_path).unwrap();
             unsafe { libc::fopen(path.as_ptr(), mode.as_ptr()) }
         };
-        ncurses::newterm(None, output, input);
+        let _ = ncurses::newterm(None, output, input).unwrap();
```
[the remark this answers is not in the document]
correabuscar: is missing a `)` here, fixing...

cursive/src/backends/curses/n.rs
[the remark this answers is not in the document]
gyscos: I meant to make `text` be a `Cow<'a, str>` by removing the `&` here, and updating the usages of this variable to deref as needed.

cursive/src/backends/curses/n.rs
```diff
@@ -393,22 +399,52 @@ impl backend::Backend for Backend {
...
             while dupes_left > 0 {
-                ncurses::addstr(text);
+                let _ = ncurses::addnstr(text, len);
                 dupes_left -= 1;
             }
         }
     }
 }
 
+#[inline]
```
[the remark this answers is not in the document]
correabuscar: roger that :)

cursive/src/backends/curses/n.rs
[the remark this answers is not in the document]
correabuscar: To be clear, when you make a suggestion, you don't really mean for me to commit it as such? even though that commit includes `Co-authored-by: Alexandre Bury <email here>` ? because then I could just go back to the original commits and mod them there(and I can add the Co-authored-by line manually) and force push, else we're piling up commits here :)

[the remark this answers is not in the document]
gyscos: Yeah I don't really care about how the change makes it in - it's convenient to apply it from github's UI (and we can squash the commits when merging), but it's also totally fine to do it separately.

cursive/src/backends/curses/n.rs
```diff
@@ -110,7 +111,12 @@ impl Backend {
             let path = CString::new(output_path).unwrap();
             unsafe { libc::fopen(path.as_ptr(), mode.as_ptr()) }
         };
-        ncurses::newterm(None, output, input);
+        ncurses::newterm(None, output, input).map_err(|_| io::Error::new(io::ErrorKind::Other, "could not call newterm"))?;
```
correabuscar, line 114: Shouldn't we do something like, keep the original error also here?
```rust
ncurses::newterm(None, output, input)
    .map_err(|e| io::Error::new(io::ErrorKind::Other, format!("could not call newterm: {}", e)))?;
```

So, instead of this:
```rust
called `Result::unwrap()` on an `Err` value: Custom { kind: Other, error: "could not call newterm" }
```

we'd get this:
```rust
called `Result::unwrap()` on an `Err` value: Custom { kind: Other, error: "could not call newterm nul byte found in provided data at position: 0" }
```
(but of course it would be a different future error(none implemented yet), this is just because I modded the arg, to actually error anything)

cursive/src/backends/curses/n.rs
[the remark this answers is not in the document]
correabuscar: So, the reason I used `&` instead of two `.as_ref()` below, was because I figured if I take the ref only once instead of twice is better. And it seemed to me it would still have the same `'a` lifetime. But I can mod to do it with two `.as_ref()` calls which won't probably make any difference(inlining?), I've no idea.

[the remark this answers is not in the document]
gyscos: I think you can just use `&text` (as suggested by the compiler in the error) rather than `.as_ref()`.
Using `&delete_nuls(text)` relies on "temporary lifetime extension", which, while valid, is a bit less obvious and not necessary here.

cursive/src/backends/curses/n.rs
```diff
@@ -110,7 +111,8 @@ impl Backend {
             let path = CString::new(output_path).unwrap();
             unsafe { libc::fopen(path.as_ptr(), mode.as_ptr()) }
         };
-        ncurses::newterm(None, output, input);
+        ncurses::newterm(None, output, input).map_err(|e| io::Error::new(io::ErrorKind::Other, format!("could not call newterm: {}",e)))?;
```
gyscos, line 114: Nit: you can directly include the variable name in the `format!` expression:
```suggestion
        ncurses::newterm(None, output, input).map_err(|e| io::Error::new(io::ErrorKind::Other, format!("could not call newterm: {e}")))?;
```

correabuscar:
I'm not sure how to disable those checks above which seem to happen after every push, but the PR is not supposed to work/compile until this `ncurses-rs` PR gets in: https://github.com/jeaye/ncurses-rs/pull/218 so it's a waste of resources to even try to 'check'

cursive-core/src/buffer.rs
```diff
@@ -183,10 +183,17 @@ impl PrintBuffer {
         // TODO: Use some WithWidth(&str, usize) to not re-compute width a thousand times
         for g in text.graphemes(true) {
             let width = g.width();
+            //if (width == 0) || ("\0" == g) {
             if width == 0 {
                 continue;
             }
-            self.set_cell(pos, g, CellWidth::from_usize(width), style);
+            if "\0" == g {
```
gyscos, line 190: We might want to do this for all control characters, not just \0.

correabuscar: Sounds good but I don't think I'm equipped to do this myself, and even do it close to properly, that's why I thought it's best to leave this for you to do(better,faster and without wasting your time(reviewing my coding mistakes) too). This was just a proof of concept commit which I can easily undo. What do you think?

Here's the list of all(?) control characters: https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=\p{Cc} taken from https://github.com/unicode-rs/unicode-width/pull/45/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759L59

Kind of interesting to see how they do it in `unicode-width`, for `&str` and for `char`:

https://github.com/unicode-rs/unicode-width/commit/4efb1803faa054f1bea3c0457275ad3c8610170b#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R201-R228

https://github.com/unicode-rs/unicode-width/commit/4efb1803faa054f1bea3c0457275ad3c8610170b#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R130-R149

cursive-core/src/buffer.rs
```diff
@@ -179,14 +179,32 @@ impl PrintBuffer {
...
         let mut pos = start;
 
+        fn is_control_char(g: &str) -> bool {
+            g.chars().all(|c| {
+                (
+                    c <= '\u{001F}'
+                    //&& (c != '\u{0009}' && c != '\u{000A}') //TODO: special case for tab or newline?
+                ) || (c >= '\u{007F}' && c <= '\u{009F}')
+            })
+        }
```
gyscos, line 189: I think specifically to check if it's a control char, we could be simpler and directly check the bytes: the bytes should be in the range `0x00..=0x1F | 0x7F..=0x9F`. This excludes any utf-8 continuation byte, so we already exclude any multi-byte character.

I thought initially that `g` would also be single-byte for control chars, but that's maybe not true (for example `\r\n` might be a single grapheme, and would still count as control char).

We should return `true` for newlines, tabs and such - basically anything that doesn't neatly fit in a single cell should return true here (and ideally should never be sent in the first place, but that's a topic for another day).

gyscos: Looks great otherwise! We're just waiting for ncurses itself to be updated, right?

correabuscar: > Looks great otherwise! We're just waiting for ncurses itself to be updated, right?

yes, but this behavior doesn't depend on ncurses being updated at all, that is, it should actually be a separate PR and be applied on the current cursive to fix #780 else it just exits currently due to a panic:
```
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s
     Running `target/x86_64-unknown-linux-gnu/debug/cursive_nul_in_text`
thread 'main' panicked at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ncurses-5.101.0/src/lib.rs:65:29:
called `Result::unwrap()` on an `Err` value: NulError(0, [0])
stack backtrace:
   0: rust_begin_unwind
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panicking.rs:72:14
   2: core::result::unwrap_failed
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/result.rs:1649:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/result.rs:1073:23
   4: <&str as ncurses::ToCStr>::to_c_str
             at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ncurses-5.101.0/src/lib.rs:65:9
   5: ncurses::addstr
             at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ncurses-5.101.0/src/lib.rs:99:23
   6: <cursive::backends::curses::n::Backend as cursive_core::backend::Backend>::print
             at /tmp/cursive/cursive/src/backends/curses/n.rs:401:9
   7: cursive_core::buffer::PrintBuffer::flush
             at /tmp/cursive/cursive-core/src/buffer.rs:291:13
   8: cursive_core::cursive_run::CursiveRunner<C>::draw
             at /tmp/cursive/cursive-core/src/cursive_run.rs:98:9
   9: cursive_core::cursive_run::CursiveRunner<C>::refresh
             at /tmp/cursive/cursive-core/src/cursive_run.rs:193:9
  10: cursive_core::cursive_run::CursiveRunner<C>::run
             at /tmp/cursive/cursive-core/src/cursive_run.rs:233:9
  11: cursive_core::cursive_root::Cursive::try_run_with
             at /tmp/cursive/cursive-core/src/cursive_root.rs:898:9
  12: <cursive_core::cursive_root::Cursive as cursive::cursive_ext::CursiveExt>::run_ncurses
             at /tmp/cursive/cursive/src/cursive_ext.rs:87:9
  13: <cursive_core::cursive_root::Cursive as cursive::cursive_ext::CursiveExt>::run
             at /tmp/cursive/cursive/src/cursive_ext.rs:76:17
  14: cursive_nul_in_text::main
             at ./src/main.rs:9:5
  15: core::ops::function::FnOnce::call_once
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

I kind of confused things by adding this commit to this unrelated PR, I shouldn't have. If this unrelated PR(about ncurses v6) would've been applied (and ncurses been updated too) then the behavior here with `\0`(which should be in a different PR, just to be clear) would only fix an issue where the size and shadows were off:
with_ncurses_6

gyscos: I cherry-picked the last commit (just removed the `\0` width test, so we're not tied to unicode-width 1.1.13)

correabuscar:
Now this PR is clean(ish), we're waiting for one of these two `ncurses-rs` PRs to get in:
* https://github.com/jeaye/ncurses-rs/pull/218 (if building properly on more than just fedora/ubuntu/nixos is desired)
* or just the bare minimum PR: https://github.com/jeaye/ncurses-rs/pull/220

Keeping the PR in draft until then. Maybe even have the optional pancurses one([PR](https://github.com/ihalila/pancurses/pull/93)) get in, even though cursive works without that one in.

PS: that [CI failure](https://github.com/gyscos/cursive/actions/runs/9403444028/job/25900009802?pr=778) is due to `TERM` not being set or something(that PR https://github.com/jeaye/ncurses-rs/pull/218 ) would actually tell us about more precisely.

cursive-core/src/buffer.rs
gyscos: > because it's now only testing that the replacement char has width 1.

Which is fine: we only print that _if_ the control char has a width of 1. So:
* On old unicode-width, control chars are ignored, no replacement.
* On new unicode-width, control chars are width=1, replaced with width=1 replacement char.

To be extra robust we could check for a double-width control char (not possible today, but maybe in a future unicode-width update? :( ) and send more replacement chars (or a double-width one) in that case. Not super important though as it's not a likely scenario.

> maybe printing the control char raw like [that](https://github.com/gyscos/cursive/commit/0673fbcbed8acb78f2d1c9a11c5158b5a621e737#diff-5ab05e635edf74924d76fd94ec7bd7712e5193641e3b12b77cc47662f6d33b64R205) might not be a good idea, if it's '\r' it would jump to start of line and overwrite that part of the panic message maybe?

Ah right, making it into a debug-print instead :)

Fixed the other one too.

gyscos:
> PS: that [CI failure](https://github.com/gyscos/cursive/actions/runs/9403444028/job/25900009802?pr=778) is due to TERM not being set or something(that PR https://github.com/jeaye/ncurses-rs/pull/218 would actually tell us about it or any other(but not all) weird errors more precisely).

Do you mean the `TERM` variable is not set _at compile time_? I think that shouldn't be required - it's not uncommon to build without a terminal at all (from CI, or even from some IDEs), and lying about a `TERM` doesn't sound like a better option.

correabuscar:
> Do you mean the `TERM` variable is not set _at compile time_? I think that shouldn't be required - it's not uncommon to build without a terminal at all (from CI, or even from some IDEs), and lying about a `TERM` doesn't sound like a better option.

Sounds reasonable to not need it, but it's true that because at `build.rs` time in `ncurses-rs`, some piece of code (that someone else made and I just carried over without giving it any thought, if I'm being honest here) will attempt to init ncurses to find whether some macros are defined. Well, it's right here: https://github.com/jeaye/ncurses-rs/blob/3aa22bc279e4929e3ab69d49f75a18eda3e431e9/src/genconstants.c#L9

correabuscar:
Closing this to allow someone else in the future to come along and do things better whenever things are ready in `ncurses-rs`, and also because I've already closed the prerequisite PR for `ncurses-rs` and without it this PR has no point:)
Also, I may or may not be here in the future to provide support for these changes, as I had originally intended.

Thank you so much @ gyscos for the reviews and replies that helped me understand rust code better! Very much appreciated! All the best!

correabuscar:
Reopening because https://github.com/jeaye/ncurses-rs/pull/220 got in, so this PR might work now, bare with me while I re-test everything in OP...

correabuscar:
I don't have the virtual machines anymore, so I've only tested Gentoo... 
PR uses v6.0.1 of `ncurses` crate which was just published.

gyscos:
Good job here, and glad it paid off :)

gyscos (approved):

PR #789, opened 2024-06-07
Tests support control chars widths 0 or 1

### LINKED ISSUE #788: [BUG] tests fail with unicode-width 0.1.12 but not with 0.1.13 (should we support both in tests too?)
**Describe the bug**

**tl;dr:** some tests still fail for users that had old `git clone` and didn't `cargo update` after `git pull`, even though the lib works as intended.
But since the lib can handle control chars of either width 0 or 1, should I mod the currently failing tests to handle both width `0` or `1` ? and should I do it without introducing a `build.rs` ? or just do nothing and thus allow tests to catch when width is `!= 1` in the future ?

------------
In this commit:
https://github.com/gyscos/cursive/commit/af7c3d7b5a5ce65ef90a12260a8f7c6477fa6c20

I see the intention was as you've [mentioned](https://github.com/gyscos/cursive/pull/778#discussion_r1629703607):
> I cherry-picked the last commit (**just removed the \0 width test, so we're not tied to unicode-width 1.1.13**)

in other words, you wanted it to work with either 0.1.12 or 0.1.13 (but also **for tests to not fail as well**), as you've [mentioned](https://github.com/gyscos/cursive/issues/780#issuecomment-2152773726) here also:
> The crash should now be fixed with the latest commit, which **ignores \0 (when using unicode-width < 0.1.13), or replaces it with � (when using unicode-width >= 0.1.13).**

however, with unicode-width 0.1.12, all control chars have width 0 (I've just tested) therefore this whole test(which already covers `\0` as well) will fail:
https://github.com/gyscos/cursive/blob/332a6a4bff93f9b2d5e9540e9cc3af43e42e636e/cursive-core/src/utils/lines/spans/tests.rs#L32-L61
so not just the test for `\0` that got removed in the aforementioned commit.

I'm not sure(unless add `build.rs`[^1]) how to conditionally make the test depend on 0.1.12 and check for width 0, and also if >=0.1.13 check for width 1, but **the initial intent was that if the assumed `width` changes in a future `unicode-width` update, we should have tests fail**, else we wouldn't know that bugs creep in due to width being different now (such as: sizes and shadows being off), even though the lib can current handle both widths: 0 or 1. Maybe given this, can mod the test for `width` being `< 2` but there's the test below which might need some tweaking too(if `build.rs` introduction isn't wanted).

Alternatively, maybe it doesn't matter in this case that the `cargo test` will fail for users which just `git pull`-ed new changes(due to them having their own `Cargo.lock` locally, because it's in `.gitignore`), because, while this test fails, the lib does work as expected, and the user is in fact expected to `cargo update` anyway.

Additionally this test will fail too, for those users that didn't `cargo update` to get unicode-width 0.1.13: `$ cargo test --example select_test -- --nocapture`, in particular `tests::control_chars_become_replacement_char` and `tests::nuls_become_replacement_char`.

But again, all the mentioned tests only fail for users that didn't `cargo update`, not for anyone who did a fresh `git clone` for example. And we kinda want these tests to fail to catch future `width` changes, I'd think.

[^1]: without introducing a `build.rs`(which would set an expected `width`(based on unicode-width version, so it would be either `0` or `1`) in a new `.rs` that we'd `include!()` and thus can use that in the test). I'd be happy to do it, if you'd just let me know that you want this.

**To Reproduce**
1. `git clone`
2. force use `unicode-width` 0.1.12
```diff
diff --git a/cursive/Cargo.toml b/cursive/Cargo.toml
index 3a26f0b..810edc0 100644
--- a/cursive/Cargo.toml
+++ b/cursive/Cargo.toml
@@ -19,7 +19,7 @@ cursive_core = { path = "../cursive-core", version= "0.3.0"}
 crossbeam-channel = "0.5"
 cfg-if = "1"
 unicode-segmentation = "1"
-unicode-width = "0.1"
+unicode-width = "=0.1.12"
 lazy_static = "1"
 libc = "0.2"
 maplit = { version = "1.0", optional = true }
```
3. `cargo update`
4. `cargo test test_control_chars_have_width_1 -- --nocapture`
5. `cargo test --example select_test -- --nocapture`

**Expected behavior**
for users with stale `Cargo.lock`, tests wouldn't fail.

**Screenshots**
N/A

**Environment**
* Operating system used
* Backend used: ncurses (the default one)
* Current locale (run `locale` in a terminal)
```
LANG=en_US.utf8
LC_CTYPE="en_US.utf8"
LC_NUMERIC="en_US.utf8"
LC_TIME="en_US.utf8"
LC_COLLATE="en_US.utf8"
LC_MONETARY="en_US.utf8"
LC_MESSAGES="en_US.utf8"
LC_PAPER="en_US.utf8"
LC_NAME="en_US.utf8"
LC_ADDRESS="en_US.utf8"
LC_TELEPHONE="en_US.utf8"
LC_MEASUREMENT="en_US.utf8"
LC_IDENTIFICATION="en_US.utf8"
LC_ALL=
```
* Cursive version (from crates.io, from git, ...)
latest commit which currently is: https://github.com/gyscos/cursive/commit/332a6a4bff93f9b2d5e9540e9cc3af43e42e636e

**Additional context**
nits: I forgot to mention that in [commit](https://github.com/gyscos/cursive/commit/0673fbcbed8acb78f2d1c9a11c5158b5a621e737#diff-5ab05e635edf74924d76fd94ec7bd7712e5193641e3b12b77cc47662f6d33b64R196-R201) in code comments, it's `unicode-width` instead of `unicode-segmentation`(which is at `1.11.0` and has no effect on width, from what I can tell)

correabuscar:
closes #788 

fixed a couple of nits too

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -24,4 +24,4 @@ jobs:
     - name: Run tests
       run: >
         cargo test --features "toml markdown ansi termion-backend crossterm-backend" --no-default-features --verbose &&
-        cargo test --example select_test -- --nocapture
+        cargo test --example select_test
--- a/cursive-core/src/buffer.rs
+++ b/cursive-core/src/buffer.rs
@@ -193,12 +193,12 @@ impl PrintBuffer {
             let width = g.width();
             if width == 0 {
                 // Any zero-width grapheme can be ignored.
-                // With unicode_segmentation < 0.1.13, this includes control chars.
+                // With unicode-width < 0.1.13, this includes control chars.
                 continue;
             }
 
             if is_control_char(g) {
-                // With unicode_segmentation >= 0.1.13, control chars have non-zero width (in
+                // With unicode-width >= 0.1.13, control chars have non-zero width (in
                 // practice width = 1).
                 debug_assert_eq!(
                     1, width,
--- a/cursive-core/src/utils/lines/spans/tests.rs
+++ b/cursive-core/src/utils/lines/spans/tests.rs
@@ -29,7 +29,7 @@ fn test_replacement_char_has_width_1() {
 }
 
 #[test]
-fn test_control_chars_have_width_1() {
+fn test_control_chars_have_width_0_or_1() {
     use unicode_width::UnicodeWidthStr;
     let control_chars = [
         "\u{0000}", "\u{0001}", "\u{0002}", "\u{0003}", "\u{0004}", "\u{0005}", "\u{0006}",
@@ -52,10 +52,9 @@ fn test_control_chars_have_width_1() {
             "it's supposed to be a string of 1 char"
         );
         let unicode_escape = format!("\\u{{{:04X}}}", c.chars().last().unwrap() as u32);
-        assert_eq!(
-            width, 1,
-            "Width of control character {} is not 1",
-            unicode_escape
+        assert!(
+            (0..=1).contains(&width),
+            "Width of control character {unicode_escape} is not 0 or 1, it's {width}"
         );
     }
 }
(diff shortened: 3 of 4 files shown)


gyscos:
Thanks for the PR!

I just pushed a change that only marks some backends "persistent". The puppet backend defaults to non-persistent.
In the print buffer, we now only skip similar chars on persistent backends. What this means is that for the pupper backend, we will flush the entire screen and not just the diffs, so we shouldn't have to use the lowercase `a` workaround.

gyscos:
Whoops indeed the last commit wasn't pushed :-'

correabuscar:
I'll update in a sec, testing and such :)

gyscos (approved):

PR #790, opened 2024-06-07
Allow finish game in example `mines`

xilec:
Allow finish game in example `mines`.

Added:
- detection of victory.
- showing mines, at the end of game.
- generation of board after first player click, for not losing on first step.

Not implemented: reveal nearby unmarket cells by middle mouse click and count scores.
There is known bug: do not shows mines when game is defeated. But almost the same implementation for victory case works.


gyscos:
Closed in favor of the larger PR doing the same thing

PR #793, opened 2024-06-12
Support autojump in popup `SelectView`

bgilbert:
For now, only support autojump when the popup is closed, and not inside the menu popup itself.

Note that autojump fires the select callback, while popup mode previously only fired the submit callback.


gyscos (approved):

gyscos:
Thanks for the work!

PR #794, opened 2024-06-12
Support replacing angle brackets in `SelectView` popup button

bgilbert:
Allow a `SelectView` to be configured with custom decorators instead of the angle brackets around the current item label.

### DIFF
--- a/cursive-core/src/views/select_view.rs
+++ b/cursive-core/src/views/select_view.rs
@@ -15,6 +15,8 @@ use std::cmp::{min, Ordering};
 use std::sync::atomic::AtomicUsize;
 use std::sync::{Arc, Mutex};
 
+use unicode_width::UnicodeWidthStr;
+
 type SelectCallback<T> = dyn Fn(&mut Cursive, &T) + Send + Sync;
 
 /// View to select an item among a list.
@@ -72,6 +74,8 @@ pub struct SelectView<T = String> {
 
     // `true` if we show a one-line view, with popup on selection.
     popup: bool,
+    // Decorators to draw around the popup button.
+    decorators: [String; 2],
 
     // We need the last offset to place the popup window
     // We "cache" it during the draw, so we need interior mutability.
@@ -102,6 +106,7 @@ impl<T: 'static + Send + Sync> SelectView<T> {
             on_submit: None,
             align: Align::top_left(),
             popup: false,
+            decorators: ["<".to_string(), ">".to_string()],
             autojump: false,
             last_offset: Mutex::new(Vec2::zero()),
             last_size: Vec2::zero(),
@@ -165,6 +170,19 @@ impl<T: 'static + Send + Sync> SelectView<T> {
     pub fn set_popup(&mut self, popup: bool) {
         self.popup = popup;
         self.last_required_size = None;
+    }
+
+    /// Use custom decorators around the popup button instead of "<" and ">".
+    ///
+    /// Chainable variant.
+    #[must_use]
+    pub fn decorators<S: Into<String>>(self, start: S, end: S) -> Self {
+        self.with(|s| s.set_decorators(start, end))
+    }
+
+    /// Use custom decorators around the popup button instead of "<" and ">".
+    pub fn set_decorators<S: Into<String>>(&mut self, start: S, end: S) {
+        self.decorators = [start.into(), end.into()];
     }
 
     /// Sets a callback to be used when an item is selected.
@@ -784,7 +802,13 @@ impl<T: 'static + Send + Sync> SelectView<T> {
         // We'll want to show the popup so that the text matches.
         // It'll be soo cool.
         let item_length = self.items[focus].label.width();
-        let text_offset = (self.last_size.x.saturating_sub(item_length)) / 2;
+        let text_offset = self
+            .last_size
+            .x
+            .saturating_sub(self.decorators_width())
+            .saturating_sub(item_length)
+            / 2
+            + self.decorators[0].width();
         // The total offset for the window is:
         // * the last absolute offset at which we drew this view
         // * shifted to the right of the text offset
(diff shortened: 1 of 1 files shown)


gyscos:
Hi, and thanks for the work!

For buttons, the user can entirely replicate the brackets from the `new_raw` function, but including the desired replacement (for example if they want square brackets, nothing at all, or even putting back the original brackets themselves!) in the given string.

Here though, what ends up being displayed is a combination of the selected entry (which can change, and should not include brackets directly) and the "decorators" (currently hard-coded to be angled brackets).

I would be fine adding an option to replace the angled brackets with user-defined decorators (potentially an empty string). I guess it would take store a `String` for each side.

bgilbert:
Good point, updated!

cursive-core/src/views/select_view.rs
```diff
@@ -965,13 +981,16 @@ impl<T: 'static + Send + Sync> View for SelectView<T> {
             printer.with_style(style, |printer| {
                 // Prepare the entire background
                 printer.print_hline((1, 0), x, " ");
-                // Draw the borders
-                printer.print((0, 0), "<");
-                printer.print((x, 0), ">");
+                // Draw the decorators
+                printer.print((0, 0), &self.decorators[0]);
+                printer.print((x + 1 - self.decorators[1].len(), 0), &self.decorators[1]);
```
gyscos, line 986: We should use `.width()` rather than `.len()` when the display size is what matters:
```suggestion
                printer.print((x + 1 - self.decorators[1].width(), 0), &self.decorators[1]);
```

cursive-core/src/views/select_view.rs
```diff
@@ -965,13 +981,16 @@ impl<T: 'static + Send + Sync> View for SelectView<T> {
...
 
                 if let Some(label) = self.items.get(focus).map(|item| &item.label) {
                     // And center the text?
-                    let offset = HAlign::Center.get_offset(label.width(), x + 1);
+                    let offset = HAlign::Center.get_offset(
+                        label.width(),
+                        x + 1 - self.decorators.iter().map(|d| d.len()).sum::<usize>(),
```
gyscos, line 992: Same here: what we want is the width (in cells), not the length (in bytes).
```suggestion
                        x + 1 - self.decorators.iter().map(|d| d.width()).sum::<usize>(),
```

gyscos: And actually let's use saturating_sub since we don't know how wide the decorators are.
```suggestion
                        (x + 1).saturating_sub(self.decorators.iter().map(|d| d.width()).sum::<usize>()),
```

cursive-core/src/views/select_view.rs
```diff
@@ -1027,7 +1046,10 @@ impl<T: 'static + Send + Sync> View for SelectView<T> {
             .max()
             .unwrap_or(1);
         let size = if self.popup {
-            Vec2::new(w + 2, 1)
+            Vec2::new(
+                w + self.decorators.iter().map(|d| d.len()).sum::<usize>(),
```
gyscos, line 1050:
```suggestion
                w + self.decorators.iter().map(|d| d.width()).sum::<usize>(),
```

gyscos (commented):
Looks good, thanks!

Just a few comments (mostly len vs width). Also since you get the sum of the decorators width, it may make sense to have a method to get it easily.

cursive-core/src/views/select_view.rs
gyscos: It may also make more sense to directly compute `x` differently above:

```rust
            let x = match printer.size.x.checked_sub(self.decorators[1].width()) {
                Some(x) => x,
                None => return,
            };
```

cursive-core/src/views/select_view.rs
gyscos: And if you compute `x = printer.size.x.checked_sub(self.decorators[1].width())`, then you only need `x.saturating_sub(self.decorators[0].width())`.

cursive-core/src/views/select_view.rs
bgilbert: Looking at this further, the use of `x` was already a bit buggy: we were trying to avoid redundantly printing under the decorators in `print_hline()` but only avoided one of them, and I think we were willing to overwrite a decorator if space-constrained by one character.  To clean this up a bit, I've redefined `x` to `lw`, the available label width after leaving space for the decorators.

bgilbert:
Updated, added a helper method to compute the total decorator width, and fixed coordinate calculation for the popup.

cursive-core/src/views/select_view.rs
```diff
@@ -957,21 +985,22 @@ impl<T: 'static + Send + Sync> View for SelectView<T> {
...
-                // Prepare the entire background
-                printer.print_hline((1, 0), x, " ");
-                // Draw the borders
-                printer.print((0, 0), "<");
-                printer.print((x, 0), ">");
+                // Prepare the label background
+                printer.print_hline((self.decorators[0].width(), 0), lw, " ");
+                // Draw the decorators
+                printer.print((0, 0), &self.decorators[0]);
+                printer.print((self.decorators[0].width() + lw, 0), &self.decorators[1]);
```
gyscos, line 998: Can we store `self.decorators[0].width()` in a variable, so we only need to compute it once in this function call?
We'll still compute it an extra time as part of `decorators_width()`, but it might sacrifice some readability to avoid that.

cursive-core/src/views/select_view.rs
```diff
@@ -957,21 +985,22 @@ impl<T: 'static + Send + Sync> View for SelectView<T> {
                 PaletteStyle::Primary
             };
 
-            let x = match printer.size.x.checked_sub(1) {
-                Some(x) => x,
+            let lw = match printer.size.x.checked_sub(self.decorators_width()) {
```
gyscos, line 988: Let's call it `available`, since that's the available width we have for the label.

bgilbert:
Done!

gyscos (approved):
Thanks for bearing with me!

bgilbert:
No problem, thanks for the reviews!

PR #795, opened 2024-06-18
Add inner getters to `FocusTracker`

bgilbert:


gyscos:
Hi, and thanks for the PR!

Indeed, looks like we forgot to include these.

gyscos (approved):

PR #796, opened 2024-06-19
make Theme::load_toml public

therealolie:
im keeping style configs together with other stuff, this allows me to not parse the config twice


gyscos (approved):
Hi, and thanks for the PR!

Looks good, let's make this public!

PR #798, opened 2024-06-20
tentatively fixes #776

### LINKED ISSUE #776: [BUG] Panic when opening empty menubar subtree
**Describe the bug**
When creating a menubar with an empty subtree, panics when you select it in the menu.
Dealing with this as I want to create a dynamic menubar to select a file, although maybe this isn't the best approach.

**To Reproduce**
```rs
let mut siv = cursive::default();
siv.menubar().add_subtree("foo", Tree::new());
siv.add_global_callback(Key::Esc, |s| s.select_menubar());
siv.run();
// Then press escape, go to foo and press enter.
```

**Expected behavior**
Expected to not panic

**Environment**
* OS: Linux (Nobara 39)
* Backend used: ncurses
* Current locale: en_US
* Cursive version: 0.20.0

**Additional context**
Assuming index OOB error "len is 0 but index is 0":
https://github.com/gyscos/cursive/blob/0a95c82c88c8a791d7fd983e7fb9f568b77551a8/cursive-core/src/views/menu_popup.rs#L331

As panic points to around there (line 335 specifically).

correabuscar:
or perhaps there's a better/different way?
#776

### DIFF
--- a/cursive-core/src/views/menu_popup.rs
+++ b/cursive-core/src/views/menu_popup.rs
@@ -130,7 +130,11 @@ impl MenuPopup {
             } else if cycle {
                 // Only cycle once to prevent endless loop
                 cycle = false;
-                self.focus = self.menu.children.len() - 1;
+                let len = self.menu.children.len();
+                if len == 0 {
+                    break;
+                }
+                self.focus = len - 1;
             } else {
                 break;
             }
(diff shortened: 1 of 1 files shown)


gyscos:
I think we could exit early from these functions if we notice at the start that `self.children.is_empty()` - there is no need to scroll, and there is no event we could react to.

correabuscar:
I think I avoided early exit via 'if' check to avoid paying the check for every normal non-empty case, which should be most of the cases. At least, that's what I told myself at the time :)


feels kinda hacky either way, maybe I wait for someone else to do it better

PR #803, opened 2024-11-10
Fixed issue 143

razvaniliescu:
Added a new parameter for the Dialog structure to enable vertical orientation of the buttons and modified the draw methods accordingly.

### DIFF
--- a/cursive-core/src/views/dialog.rs
+++ b/cursive-core/src/views/dialog.rs
@@ -565,7 +565,7 @@ impl Dialog {
         match result {
             EventResult::Ignored => {
                 match event {
-                    // Up goes back to the content
+                    // Up goes back to the content if buttons are horizontal
                     Event::Key(Key::Up) if self.vertical_button_orientation == false => {
                         if let Ok(res) = self.content.take_focus(Direction::down()) {
                             self.focus = DialogFocus::Content;
@@ -575,6 +575,7 @@ impl Dialog {
                         }
                     }
 
+                    // Left goes back to the content if buttons are vertical
                     Event::Key(Key::Left) if self.vertical_button_orientation == true => {
                         if let Ok(res) = self.content.take_focus(Direction::down()) {
                             self.focus = DialogFocus::Content;
@@ -647,7 +648,7 @@ impl Dialog {
             false => {
                 let mut buttons_height = 0;
                 // Current horizontal position of the next button we'll draw.
-        
+                
                 // Sum of the sizes + len-1 for margins
                 let width = self
                     .buttons
@@ -692,6 +693,7 @@ impl Dialog {
                 return Some(buttons_height);
             }
             true => {
+                // Current horizontal position of the next button
                 let mut buttons_width = 0;
 
                 // Calculate the total height of buttons including spacing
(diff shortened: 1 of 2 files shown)


cursive-core/src/views/dialog.rs
```diff
@@ -613,69 +644,148 @@ impl Dialog {
...
-            None => return,
-        };
-
-        self.content.draw(
-            &printer
-                .offset(self.borders.top_left() + self.padding.top_left())
-                .cropped(inner_size)
-                .focused(self.focus == DialogFocus::Content),
-        );
+        match self.vertical_button_orientation {
```
gyscos, line 749: Here too we should be able to share the implementation.

Either directly take `buttons_size` as a `Vec2` in the function, or:
```rust
let buttons_size = Vec2::from_major_minor(self.buttons_orientation, buttons_size, 0)`
```

Then you can do

```rust
let taken = buttons_size + self.borders.combined() + self.padding.combined();
```

And continue with the shared logic.

cursive-core/src/views/dialog.rs
```diff
@@ -745,86 +855,186 @@ impl Dialog {
 
 impl View for Dialog {
     fn draw(&self, printer: &Printer) {
-        // This will be the buttons_height used by the buttons.
-        let buttons_height = match self.draw_buttons(printer) {
-            Some(height) => height,
-            None => return,
-        };
+        match self.vertical_button_orientation {
```
gyscos, line 858: Same here, we should be able to abstract over the orientation and have a single implementation.

gyscos (commented):
Thank you for the work!

I think we should be able to keep a shared implementation in most cases, using various helper methods around `Orientation`. Let me know if you need some help!

cursive-core/src/views/dialog.rs
```diff
@@ -555,15 +565,26 @@ impl Dialog {
...
+                            self.focus = DialogFocus::Content;
+                            res
+                        } else {
+                            EventResult::Ignored
+                        }
+                    }
+
+                    // Left goes back to the content if buttons are vertical
+                    Event::Key(Key::Left) if self.vertical_button_orientation == true => {
                         if let Ok(res) = self.content.take_focus(Direction::down()) {
```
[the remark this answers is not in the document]
razvaniliescu: Changed Direction::down to Direction::right

cursive-core/src/views/dialog.rs
```diff
@@ -555,15 +565,26 @@ impl Dialog {
         match result {
             EventResult::Ignored => {
                 match event {
```
gyscos, line 567: You could `match (event, self.button_orientation) {`, then match on `(Event::Key(Key::Up), Orientation::Horizontal)`, which could be easier than adding `if self.button_orientation == ...` every time.

razvaniliescu: Done

cursive-core/src/views/dialog.rs
razvaniliescu: I managed to turn it into a single implementation with the advice you provided.

cursive-core/src/views/dialog.rs
```diff
@@ -613,69 +644,148 @@ impl Dialog {
...
-                .h
-                .get_offset(width, printer.size.x - overhead.horizontal());
-
-        let overhead_bottom = self.padding.bottom + self.borders.bottom + 1;
-
-        let y = match printer.size.y.checked_sub(overhead_bottom) {
-            Some(y) => y,
-            None => return None,
-        };
+        match self.vertical_button_orientation {
```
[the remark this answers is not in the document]
razvaniliescu: I remade the draw_buttons implementation using the Vec2::from_major_minor method.

cursive-core/src/views/dialog.rs
razvaniliescu: I remade the draw implementation. It has a single branch now, with some additional steps if the orientation is vertical.

razvaniliescu:
Is there anything you would like me to change in the implementation?

PR #810, opened 2025-02-22
Move to use `RefCell` in `Printer`

xilec:
`Printer` already is `!Sync` therefore it seems more suitable use `&RefCell<PrintBuffer>` instead of `&RwLock<PrintBuffer>`. It will be much more convenient for debug (panic instead of deadlock in case of incorrect use) and faster.


xilec:
@gyscos, can you review this PR?

PR #814, opened 2025-03-21
Add LICENSE to cursive-macros

xanderlent:
As with gyscos/cursive#702, symlink the LICENSE into cursive-macros. This allows cargo to bundle it into the crate.

The LICENSE text is needed by Linux distributions, among others.


gyscos (approved):
Thanks for the work!

PR #817, opened 2025-04-05
Markdown fixes

basebox-mt:
This PR contains Markdown usage samples in the documentation and the examples folder as well as some opinionated fixes to the Markdown handling itself:

* All headings are rendered in bold font, no matter what level
* Blockquotes are rendered in italic with a single leading `>` for easier soft line breaks
* Never allows more than one blank line
* Inserts one blank line before and after headings, code blocks, blockquotes

The parsing test has been changed to make it easier to find the failing span; it has also been extended to test the new behaviour.

`pulldown-cmark` seems to have a bug that always returns an empty title for links. I updated it to version 0.13 in Cargo.toml, but this unfortunately did not change this.

### DIFF
--- a/cursive-core/src/utils/markup/markdown.rs
+++ b/cursive-core/src/utils/markup/markdown.rs
@@ -1,6 +1,30 @@
 //! Parse markdown text.
 //!
 //! Needs the `markdown` feature to be enabled.
+//!
+//! ### Examples
+//!
+//! ```rust
+//! use cursive::utils::markup::parse;
+//! use cursive::views::TextView;
+//! use cursive::{Cursive, CursiveExt};
+//!
+//! let mut siv = Cursive::default();
+//!
+//! let content = parse(
+//!     r#"# Example Markdown
+//! Hello, *world*!
+//!
+//! ## Header 2
+//!
+//! * List item 1
+//! * List item 2
+//! "#,
+//! );
+//!
+//! siv.add_layer(TextView::new(content));
+//! siv.run();
+//! ```
 #![cfg(feature = "markdown")]
 #![cfg_attr(feature = "doc-cfg", doc(cfg(feature = "markdown")))]
 
--- a/cursive-core/src/utils/markup/markdown.rs
+++ b/cursive-core/src/utils/markup/markdown.rs
@@ -5,7 +5,7 @@
 //! ### Examples
 //!
 //! ```rust
-//! use cursive::utils::markup::parse;
+//! use cursive::utils::markup::markdown::parse;
 //! use cursive::views::TextView;
 //! use cursive::{Cursive, CursiveExt};
 //!
@@ -34,10 +34,16 @@ use crate::style::{Effect, Style};
 use crate::utils::markup::{StyledIndexedSpan, StyledString};
 use crate::utils::span::IndexedCow;
 
-use pulldown_cmark::{self, CowStr, Event, Tag, TagEnd};
+use pulldown_cmark::{self, CowStr, Event, LinkType, Tag, TagEnd};
 use unicode_width::UnicodeWidthStr;
 
-/// Parses the given string as markdown text.
+/// Parses the given string as Markdown text.
+///
+/// # Arguments
+/// * `input` - The markdown text to parse.
+///
+/// # Returns
+/// A `StyledString` containing the parsed text.
 pub fn parse<S>(input: S) -> StyledString
 where
     S: Into<String>,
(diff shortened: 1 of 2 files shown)


cursive-core/src/utils/markup/markdown.rs
```diff
@@ -1,19 +1,49 @@
...
 //! Parse markdown text.
 //!
 //! Needs the `markdown` feature to be enabled.
-#![cfg(feature = "markdown")]
-#![cfg_attr(feature = "doc-cfg", doc(cfg(feature = "markdown")))]
+//!
+//! ### Examples
+//!
+//! ```rust
+//! use cursive::utils::markup::markdown::parse;
```
gyscos, line 10: You might need `extern crate cursive_core as cursive`, so it really just needs cursive_core, but it looks as cursive for users.

basebox-mt: Thanks - strange though that I cannot reproduce this locally. I'll try.

gyscos: You'll need to test with the markdown feature to run these

basebox-mt: Got it, now I can at least reproduce

basebox-mt: Can't seem to fix it... it cannot resolve `CursiveExt`, and when I remove it, the compiler can't find the `run()` method.

basebox-mt: `CursiveExt` is in a different crate, this might just not be possible. If it's ok I remove the example in the doc comment. I put an Markdown example into the examples folder.

gyscos: You also don't need to _run_ the cursive app (or even make one), you can focus on parsing the markdown and building the TextView

basebox-mt:
I set the example in the doc comment to "ignore" - let me know if that's ok for you or of course feel free to change.

cursive-core/src/utils/markup/markdown.rs
```diff
@@ -76,44 +112,117 @@ impl Iterator for Parser<'_> {
...
                             heading(level as usize)
                         )));
                     }
-                    Tag::BlockQuote(_) => return Some(self.literal("> ")),
+                    Tag::BlockQuote(_) => {
+                        /* Push italic style */
+                        self.stack.push(Style::from(Effect::Italic));
+                        /* Force a blank line before a blockquote */
+                        self.new_paragraph = true;
+                        return Some(self.literal("\n\n> "));
```
gyscos: Should we still insert 2 newlines, even if this is the first thing in the document?

cursive-core/src/utils/markup/markdown.rs
```diff
@@ -76,44 +112,117 @@ impl Iterator for Parser<'_> {
...
+                    } => {
+                        self.new_paragraph = false;
+                        return match link_type {
+                            LinkType::Inline => {
+                                /* There seems to be a bug in pulldown_cmark where it doesn't
+                                 * recognize the link type. If the title is empty, we will
+                                 * render an Autlink instead of an Inline link.
+                                 */
+                                return if title.is_empty() {
+                                    Some(self.literal(format!("<{dest_url}> ")))
```
gyscos: Does it include a trailing space on purpose? If so, let's add a comment to make it clear.

cursive-core/src/utils/markup/markdown.rs
```diff
@@ -154,58 +265,127 @@ mod tests {
...
+            Span {
+                content: "<https://en.wikipedia.org> ",
+                width: "<https://en.wikipedia.org> ".width(),
+                attr: &style_none,
+            },
+            Span {
+                content: "Link",
+                width: "Link".width(),
+                attr: &style_none,
+            },
```
gyscos: Does that mean `[link](url)` is rendered `<url> link`? Is that the intended behavior?

cursive-core/src/utils/markup/markdown.rs
```diff
@@ -154,58 +265,127 @@ mod tests {
...
+            Span {
+                content: "\n\n",
+                width: "\n\n".width(),
+                attr: &style_none,
+            },
+            Span {
+                content: "\n\n",
+                width: "\n\n".width(),
+                attr: &style_none,
+            },
```
gyscos: 4 new lines at the end of the doc? Maybe we could avoid that.

PR #827, opened 2025-07-05
Fix Alt+Shift+Char for crossterm backend

azat:
Before this patch `cargo run --example key_codes --features crossterm-backend`:

    AltChar('e')
    Char('E')

After:

    AltChar('e')
    AltChar('E')


gyscos:
Thanks for the work!

PR #828, opened 2025-07-05
Properly clear terminal on Cursive::clear()

azat:
Using `buffer.write().clear()` over `printer.clear()` (which is `self.buffer.write().fill(" ")`) helps with properly clear the cursive screen area after invoking side TUI apps within it (i.e. vim)

Note, that this is not enough to clear the complete screen in case of using ShadowView.

### DIFF
--- a/cursive-core/src/cursive_root.rs
+++ b/cursive-core/src/cursive_root.rs
@@ -39,7 +39,7 @@ pub struct Cursive {
 
     menubar: views::Menubar,
 
-    pub(crate) needs_clear: bool,
+    needs_clear: bool,
 
     running: bool,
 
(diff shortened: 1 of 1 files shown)


azat:
Hi @gyscos!

I will be OK with a separate method, but AFAICS the `clear` is not a hot path, and usually called explicitly by user

gyscos:
I just tested, and there's a few main things that require clearing the screen:
* Screen is resized. If this happens it's totally fine to drop the write buffer.
* A layer is removed. In this case it's unfortunate if we need to flush the entire screen again, as it could require a very small "diff" instead.

I'd rather expose a dedicated function to call after coming back from a separate app.

gyscos (approved):
LGTM, merging.

PR #867, opened 2026-04-12
Support tuple structs in wrap_impl!

### LINKED ISSUE #866: Wrapping views in tuple structs
**Is your feature request related to a problem? Please describe.**
I would like to use the `wrap_impl!` macro with a tuple struct.

**Describe the solution you'd like**
The `wrap_impl!` macro currently only has a single branch that matches on `(self.$v:ident: $t:ty)`. If we added another branch that matches on `(self.$v:literal: $t:ty)` (possibly with some post-processing to ensure it is a non-negative integer literal).

**Describe alternatives you've considered**
I can use a standard struct with named fields.

**Additional context**
If you want to alter a small amount of behavior and don't require any additional data/configuration/etc. to do so, then you really only need to store the `View` that is being "wrapped".

For example, I wanted to make an `Unfocusable` wrapper that prevented the wrapped view from receiving focus. I was hoping to do something like this:

```rust
struct Unfocusable<V: View>(V);

impl<V: View> ViewWrapper for Unfocusable<V> {
    wrap_impl!(self.0: V);

    fn wrap_take_focus(&mut self, source: Direction) -> Result<EventResult, CannotFocus> {
        Err(CannotFocus)
    }
}
```

I'm willing to create a PR for this. Before I start working on it, though, I wanted to see if this is something you would be interested in. Maybe there's a reason to avoid this pattern that I'm missing.

---

P.S. I am open to any suggestions on my concrete idea of an `Unfocusable` wrapper, although this issue is specifically for wrapping tuple-style structs.

HarrisonMc555:
Change the fragment specifier used in `wrap_impl!` from `ident` to `tt`. Using `ident` was sufficient for "standard" structs with named fields. However, tuple structs are accessed via index numbers, which are not matched by `ident`. Although using `literal` for tuple indices seems like it should work, it does not.

Sidestep this issue by matching an `tt`, i.e. any valid token tree.

Fixes #866

### DIFF
--- a/cursive-core/src/view/view_wrapper.rs
+++ b/cursive-core/src/view/view_wrapper.rs
@@ -157,11 +157,17 @@ impl<T: ViewWrapper> View for T {
 /// impl<T: View> ViewWrapper for FooView<T> {
 ///     cursive_core::wrap_impl!(self.view: T);
 /// }
+///
+/// struct BarView<T: View>(T);
+///
+/// impl<T: View> ViewWrapper for BarView<T> {
+///     cursive_core::wrap_impl!(self.0: T);
+/// }
 /// # fn main() { }
 /// ```
 #[macro_export]
 macro_rules! wrap_impl {
-    (self.$v:ident: $t:ty) => {
+    (self.$v:tt: $t:ty) => {
         type V = $t;
 
         fn with_view<F, R>(&self, f: F) -> ::std::option::Option<R>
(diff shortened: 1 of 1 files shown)


gyscos:
Thanks for the PR!

Would you mind adding a unit test that makes sure this macro works with both a named struct and a tuple struct?

HarrisonMc555:
Sure thing! There was already a doc test, but I added a unit test. Right now it's mostly just exercising that you can invoke the macro without causing compiler errors. Let me know if there's anything else you'd like to see tested.

gyscos (approved):

gyscos:
Thanks again for the work!

PR #876, opened 2026-07-10
make cb_cink return a CbSink instead of a reference to a CbSink

Cheese-Space:
you can clone a crossbeam_channel::sender (the sender returned by the cb_cink function) cheaply, so in my opinion it would be better to not return a reference


gyscos:
Hi, and thanks for the work!

Sure, I agree the ergonomics improvement is most likely worth the tiny performance cost.

gyscos (approved):

PR #877, opened 2026-07-12
Make progress_bar inherit background

sylfn:
This fixes a TODO in the code that broke progress bars when active palette uses `Color::TerminalDefault` for backgrounds


gyscos (approved):
Thanks for the work!

